PR #699, opened 2019-03-17
Recurring donation

niteshgrg:


norbertschuler:
This is pretty old now. Will this ever be used any more? If not I suggest should close this pull request. (We can keep the branch if the code is valuable.)

PR #704, opened 2019-03-19
Feature/sync forms

niteshgrg:


norbertschuler:
This is pretty old now. Will this ever be used any more? If not I suggest should close this pull request. (We can keep the branch if the code is valuable.)

norbertschuler:
Nitesh wrote: 
In sync-forms. react-native updated version started giving out error which might be fixed now after recent update will check. This doesn’t effect anything it automates forms updation which is currently manual.

PR #1066, opened 2019-07-10
rename branch dev to test

norbertschuler:
I would like to rename the branch `dev` to `test` in our app repository to match the naming in the other repositories and the server namings. I already (1) prepared the integration job of the test branch in the TTC backend repository to support the new name by https://github.com/Plant-for-the-Planet-org/treecounter-platform/commit/1eee807445d0fa46920b6aa59ba4b8f2ce339e07 and (2) also changed the Dashboard to the new name by https://github.com/Plant-for-the-Planet-org/dashboard/commit/137ad2bcee5ee7fee1c85128f327686ece700001. (3) I created a branch named `test` on the same level as the current branch `dev`. Last thing to do would be to merge this pull request into all main branches (I also added a branch protection rule for `test` next to `dev`.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
         - master
         - staging
         - devel
-        - dev
+        - test
     docker:
       - image: alpine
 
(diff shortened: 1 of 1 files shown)


norbertschuler:
For developers this means, that everyone who has checked out `dev` would need to checkout `test` instead and rebase any local work not yet pushed onto that branch. Is anything else build upon this branch, e.g. any testing or else?

PR #1090, opened 2019-07-12
Fix unhandled rejection in getAccessToken

crucialfelix:
I think this is responsible for many of these errors:
https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5d05f81ab3cb100019330da9?filters[event.since][0]=30d&filters[error.status][0]=open

1. Always reject with an `Error` — this is how bugsnag gets it's stack trace.

2. In `getAccessToken` there can be Errors thrown inside a `.then` handler which aren't handled at all. Fortunately modern browsers and Node will actually complain about those. It used to be they were swallowed and the dev never knew anything was wrong. But that's why the bugsnag reports don't have stack traces and just say things like "Timeout"

So now if something goes wrong in there, we will get an Error (will hit the error boundary) and a stack trace.

Or, we could catch errors there and at least log them, but then:

```js
// Could catch them here and log out the user,
// but then we would never know what happened.
// Otherwise the error propagates
try {
  // then this may fail. The user is then considered logged out.
  const newToken = await refreshTokenIfExpired();
  if (newToken) {
    return newToken;
  }
} catch (error) {
  // but if we made any mistake in our code, we will never know about it
  console.log(error);
  return null;
}
```

### DIFF
--- a/app/stores/localStorage.js
+++ b/app/stores/localStorage.js
@@ -7,7 +7,7 @@ export const loadState = () => {
       }
       resolve(JSON.parse(serializedState));
     } catch (err) {
-      reject(null);
+      reject(err);
     }
   });
 };
@@ -28,7 +28,7 @@ export const fetchItem = key => {
     if (window.localStorage.getItem(key)) {
       resolve(window.localStorage.getItem(key));
     } else {
-      reject('Key not found');
+      reject(new Error(`${key} not in localStorage`));
     }
   });
 };
--- a/app/stores/localStorage.js
+++ b/app/stores/localStorage.js
@@ -25,12 +25,23 @@ export const saveItem = (key, value) => {
 
 export const fetchItem = key => {
   return new Promise(function(resolve, reject) {
-    if (window.localStorage.getItem(key)) {
-      resolve(window.localStorage.getItem(key));
-    } else {
+    const got = window.localStorage.getItem(key);
+    if (got === null) {
       reject(new Error(`${key} not in localStorage`));
+    } else {
+      resolve(got);
     }
   });
+};
+
+/**
+ * Get a possibly unset localStorage item.
+ * Use this when it is not an error for the item to be unset.
+ *
+ * @returns string | null
+ */
+export const getItem = key => {
+  return window.localStorage.getItem(key);
 };
 
 export const clearStorage = () => {
--- a/app/utils/user.js
+++ b/app/utils/user.js
@@ -1,25 +1,22 @@
-import { fetchItem, saveItem } from '../stores/localStorage';
 import jwtDecode from 'jwt-decode';
-import { postRequest } from './api';
-import { email } from '../assets';
 
-export const getAccessToken = () => {
-  return fetchItem('token')
-    .then(async token => {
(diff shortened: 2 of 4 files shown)


app/stores/localStorage.js
```diff
@@ -25,14 +25,25 @@ export const saveItem = (key, value) => {
 
 export const fetchItem = key => {
   return new Promise(function(resolve, reject) {
-    if (window.localStorage.getItem(key)) {
-      resolve(window.localStorage.getItem(key));
+    const got = window.localStorage.getItem(key);
```
crucialfelix: Minor optimization

app/utils/user.js
```diff
@@ -1,25 +1,22 @@
...
 import jwtDecode from 'jwt-decode';
+
+import { fetchItem, getItem, saveItem } from '../stores/localStorage';
 import { postRequest } from './api';
-import { email } from '../assets';
-
-export const getAccessToken = () => {
-  return fetchItem('token')
-    .then(async token => {
-      let expired = await tokenIsExpired();
```
crucialfelix: Any error that happens inside this `.then` is unhandled. I would also have thought that the `async` meant that it was like returning a `Promise` and any bad things within become a rejection.

Here's an example: https://repl.it/repls/UnwieldyGrayComputationallinguistics

app/utils/user.js
```diff
@@ -39,6 +36,22 @@ const getExpirationTimeStamp = token => {
   return getCurrentUnixTimestamp() + exp - iat;
 };
 
+/**
+ * @returns string | void
+ */
+const refreshTokenIfExpired = async () => {
+  if (await tokenIsExpired()) {
+    const prev_refresh_token = await fetchItem('refresh_token');
+    const response = await postRequest('gesdinet_jwt_refresh_token', {
```
crucialfelix: Many of the unhandled errors just say `Timeout` so it's probably a server timeout.  Many of them appear to be previously logged in users who are returning 17 days later. Then they click on My Trees and that's when the error happens.

Maybe this could be handled better here with a retry.

app/utils/user.js
```diff
@@ -1,25 +1,22 @@
...
+
+/**
+ * @returns string | null
+ */
+export const getAccessToken = async () => {
+  const token = getItem('token');
+  if (token) {
+    // This may throw an Error if localStorage is broken
+    // or POST requests timeout
+    const newToken = await refreshTokenIfExpired();
```
crucialfelix: I moved this to a separate function so it could be `try {} catch(e) {}` or it could retry.

app/utils/user.js
```diff
@@ -1,25 +1,22 @@
...
+    if (newToken) {
+      return newToken;
+    }
+  }
+  return token;
```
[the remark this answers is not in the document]
crucialfelix: ?

norbertschuler:
As I am not a experienced React developer, I cannot judge all your code, but if you are happy with it, I'd like to merge it for you.

crucialfelix:
Yes, please merge.

PR #1107, opened 2019-07-14
added notes about running iOS or Android development builds

norbertschuler:


norbertschuler:
The intention for this commit was to remove the wrong info about commenting out a line not existing any more from .babelrc

PR #1128, opened 2019-07-17
Fix/bugsnag 11 july

niteshgrg:

PR #1132, opened 2019-07-18
fixed #1125 by removing stacktrace-parser modules

### LINKED ISSUE #1125: Usage of StackTrace-Parser?
In index.native.js https://www.npmjs.com/package/stacktrace-parser is imported and the app tries to catch errors and format a stacktrace for them with this module, but then the `console.log` to print the stack trace is commented. So is this still used or can be removed?

```javascript
const parseErrorStack = error => {
  if (!error || !error.stack) {
    return [];
  }
  return Array.isArray(error.stack)
    ? error.stack
    : stacktraceParser.parse(error.stack);
};

// intercept react-native error handling
if (ErrorUtils.getGlobalHandler) {
  //console.log('ErrorUtils._globalHandler', ErrorUtils._globalHandler);
  this.defaultHandler =
    (ErrorUtils.getGlobalHandler && ErrorUtils.getGlobalHandler()) ||
    ErrorUtils._globalHandler;
  ErrorUtils.setGlobalHandler(wrapGlobalHandler); // feed errors directly to our wrapGlobalHandler function
}

async function wrapGlobalHandler(error, isFatal) {
  const stack = parseErrorStack(error);
->  //console.log(error, stack, isFatal);
  Alert.alert('Error', 'Something went wrong.');
  this.defaultHandler(error, isFatal); //after you're finished, call the defaultHandler so that react-native also gets the error
}

```

norbertschuler:
... as I think both are not used any more as we now use Bugsnag.

### DIFF
--- a/index.native.js
+++ b/index.native.js
@@ -1,5 +1,6 @@
 import { AppRegistry, Platform } from 'react-native';
 import App from './app/components/App';
+import './ReactotronConfig';
 /* app.js */
 
 console.disableYellowBox = true;
--- a/index.native.js
+++ b/index.native.js
@@ -1,36 +1,8 @@
 import { AppRegistry, Platform } from 'react-native';
 import App from './app/components/App';
-import './ReactotronConfig';
-import stacktraceParser from 'stacktrace-parser';
-import { Alert } from 'react-native';
 /* app.js */
 
 console.disableYellowBox = true;
 console.ignoredYellowBox = ['Warning: Each', 'Warning: Failed'];
-
-const parseErrorStack = error => {
-  if (!error || !error.stack) {
-    return [];
-  }
-  return Array.isArray(error.stack)
-    ? error.stack
-    : stacktraceParser.parse(error.stack);
-};
-
-// intercept react-native error handling
-if (ErrorUtils.getGlobalHandler) {
-  //console.log('ErrorUtils._globalHandler', ErrorUtils._globalHandler);
-  this.defaultHandler =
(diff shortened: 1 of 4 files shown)


index.native.js
```diff
@@ -1,36 +1,8 @@
 import { AppRegistry, Platform } from 'react-native';
 import App from './app/components/App';
-import './ReactotronConfig';
-import stacktraceParser from 'stacktrace-parser';
```
norbertschuler: stacktrace-parser was already no more included in `package.json`

norbertschuler:
Is anyone missing this code when I remove it?

index.native.js
niteshgrg: @norbertschuler We do use reactotron to inspect app network calls. It would be helpful to leave it. I have just updated it to its latest version as well in another PR.

niteshgrg: https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1128/files

norbertschuler: Ok, I won’t remove reactotron then. What about stacktrace-parser? Anyone want to keep that?

norbertschuler: > https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1128/files

@niteshgrg As I added reactotron again in the pull request, I also implicitly updated the version. Can you remove your version upgrade from https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1128/files again to avoid a later merge conflict?

PR #1137, opened 2019-07-19
Fix: Unhandled Errors thrown inside a .then() handler in getAccessToken

### LINKED ISSUE #1134: Fix unhandled rejection in getAccessToken
This PR needs to be redone:
https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1090

to add changes to `localStorage.native.js` as well.

The new PR needs to be properly tested.

Possibly add jest tests.

crucialfelix:
This fixes minor issues in `app/utils/user.js` `getAccessToken` (used for logging in and session management)

There are many reported errors coming from `/login`:

https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5d05f81ab3cb100019330da9?filters[search][0][type]=eq&filters[search][0][value]=UnhandledRejection

```
3 hours ago Aug 2nd, 20:48:44 UTC 
UnhandledRejection/login
Rejection reason was not an Error. See "Promise" tab for more detail.
rejection reason
undefined (or null)
(No stack trace)
```

This PR doesn't attempt to fix the error - it is just to supply an Error and get a stack trace so we know what the error is so it can then be fixed. 

Could be:
- localStorage failure (web or react-native), 
- API post request timeout.
- Unknown unknown

It was previously submitted and merged: https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1090 

This was missing one function in `localStorage.native.js`

@norbertschuler reverted the commit in: https://github.com/Plant-for-the-Planet-org/treecounter-app/commit/ceae9845197883ac6d5b3718e0c6b0979138951d
and then there were several other attempts (while I was on vacation) to revert, fix, reapply etc.

eg. https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1133
which was merged, then again mostly reverted.

For a minor fix, it's now become quite confusing and time consuming.

The original PR fixed:

1. Always reject with an Error — this is how bugsnag gets it's stack trace.
  That fix is now already in develop.

2. Flattened the mixed pyramids of Promise and async functions so that it's easier to follow, test and debug.

---

It is newly rebased on develop.

I tested login / logout and session refresh on web and react-native iOS (same code as for Android)

Fixes #1134  - Fix unhandled rejection in getAccessToken

### DIFF
--- a/app/utils/user.js
+++ b/app/utils/user.js
@@ -42,9 +42,17 @@ const getExpirationTimeStamp = token => {
 const refreshTokenIfExpired = async () => {
   if (await tokenIsExpired()) {
     const prev_refresh_token = await fetchItem('refresh_token');
-    const response = await postRequest('gesdinet_jwt_refresh_token', {
-      refresh_token: prev_refresh_token
-    });
+    let response;
+    try {
+      response = await postRequest('gesdinet_jwt_refresh_token', {
+        refresh_token: prev_refresh_token
+      });
+    } catch (error) {
+      // 400
+      console.log(`Failed to refresh token: ${error}`);
+      return;
+    }
+
     const newToken = response.data.token;
     const refreshToken = response.data.refresh_token;
     updateJWT(newToken, refreshToken);
--- a/app/utils/user.js
+++ b/app/utils/user.js
@@ -1,33 +1,22 @@
-import { fetchItem, saveItem } from '../stores/localStorage';
 import jwtDecode from 'jwt-decode';
-import { postRequest } from './api';
-import { email } from '../assets';
 
-export const getAccessToken = () => {
-  return fetchItem('token')
-    .then(async token => {
-      try {
-        let expired = await tokenIsExpired();
-        if (expired) {
-          const prev_refresh_token = await fetchItem('refresh_token');
-          const response = await postRequest('gesdinet_jwt_refresh_token', {
-            refresh_token: prev_refresh_token
-          });
-          const { token, refresh_token } = response.data;
-          updateJWT(token, refresh_token);
-          return token;
-        } else {
-          return token;
-        }
-      } catch (err) {
-        console.log(err);
-        return null;
-      }
-    })
-    .catch(err => {
-      console.log(err);
-      return null;
-    });
+import { fetchItem, getItem, saveItem } from '../stores/localStorage';
+import { postRequest } from './api';
+
+/**
+ * @returns string | null
+ */
+export const getAccessToken = async () => {
+  const token = await getItem('token');
+  if (token) {
+    // This may throw an Error if localStorage is broken
+    // or POST requests timeout
+    const newToken = await refreshTokenIfExpired();
+    if (newToken) {
+      return newToken;
+    }
+  }
+  return token;
 };
 
 export const updateJWT = (token, refresh_token) => {
@@ -45,6 +34,22 @@ export const updateActivateToken = (email, activate_token) => {
 const getExpirationTimeStamp = token => {
   const { iat, exp } = jwtDecode(token);
   return getCurrentUnixTimestamp() + exp - iat;
+};
+
+/**
+ * @returns string | void
+ */
+const refreshTokenIfExpired = async () => {
+  if (await tokenIsExpired()) {
+    const prev_refresh_token = await fetchItem('refresh_token');
+    const response = await postRequest('gesdinet_jwt_refresh_token', {
+      refresh_token: prev_refresh_token
+    });
+    const newToken = response.data.token;
+    const refreshToken = response.data.refresh_token;
+    updateJWT(newToken, refreshToken);
+    return newToken;
+  }
 };
 
 const tokenIsExpired = async () => {
--- a/app/stores/localStorage.js
+++ b/app/stores/localStorage.js
@@ -26,12 +26,22 @@ export const saveItem = (key, value) => {
 export const fetchItem = key => {
   return new Promise(function(resolve, reject) {
     const got = window.localStorage.getItem(key);
-    if (got) {
-      resolve(got);
-    } else {
+    if (got === null) {
       reject(new Error(`${key} not in localStorage`));
+    } else {
+      resolve(got);
     }
   });
+};
+
+/**
+ * Get a possibly unset localStorage item.
+ * Use this when it is not an error for the item to be unset.
+ *
+ * @returns string | null
+ */
+export const getItem = async key => {
+  return window.localStorage.getItem(key);
 };
 
 export const clearStorage = () => {
--- a/app/stores/localStorage.native.js
+++ b/app/stores/localStorage.native.js
@@ -35,6 +35,16 @@ export const fetchItem = async key => {
   }
 };
 
+/**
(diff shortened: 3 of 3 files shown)


norbertschuler:
@niteshgrg please check this pull request and decide if you'd prefer it to your own pull request https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1133 ?

norbertschuler:
@crucialfelix @niteshgrg 
I first merged https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1133 but reverted the commits again now after I found out staging can not react correctly on an outdated refresh-token which results into the refresh-API call `POST https://staging.trilliontreecampaign.org/api/token/refresh?version=v1.1&_locale=de` (plus a JSON payload with the refresh token inside) to fail with

```
{"code":400,"message":"Validation Failed","errors":{"errors":["Bad credentials"],"children":{"_username":{"errors":["Bad credentials"]},"_password":[]}}}
```
It also fails with the following exception if I delete the refresh-token from the local database of the browser:
```
localStorage.js:30 Uncaught (in promise) Error: token not in localStorage
    at localStorage.js:30
    at new Promise (<anonymous>)
    at t.fetchItem (localStorage.js:27)
    at user.js:10
    at tryCatch (runtime.js:65)
    at Generator._invoke (runtime.js:303)
    at Generator.e.<computed> [as next] (runtime.js:117)
    at step (bundle.js:50111)
    at bundle.js:50111
    at new Promise (<anonymous>)
```

If this patch can be made the way it's avoid above problems, that would be nice.

norbertschuler:
As login on staging server was broken, I felt forced to make a fix in the develop branch by these three commits which now conflicts with your branch: 
- aee3e41fa7ec27657e0a078a70128c9a19e845c0
- dee53f05d1361a23af8b1c18fd61df2281117fc3 (this one is just a stupid error changed by the next commit)
- ecac5a7b213c1b42665b05486e91ad7f795f0143

norbertschuler:
@crucialfelix maybe we now do not need this pull request any more. Can you check?

norbertschuler:
I will check this once more after the release 1.0.5 is done and merge if it works. I am just unsure if all rejections from promised will be caught and the code not stopped by the rejections as of now.

app/utils/user.js
```diff
@@ -7,16 +7,21 @@ import { postRequest } from './api';
...
+      // This may throw an Error if localStorage is broken
+      // or POST requests timeout
+      const newToken = await refreshTokenIfExpired();
+      if (newToken) {
+        return newToken;
+      }
     }
+    return token;
+  } catch (err) {
+    console.log(err);
```
crucialfelix: > the app shouldn't just stop on a 400 error and throw an exception, but continue to show the login

So an error happened in `refreshTokenIfExpired()` — a 400 error returned from the server. That's the problem. That problem needs to be handled there, not here.

The declared responsibility of `refreshTokenIfExpired` is to do just that "refresh token if expired". It must handle errors.

By catching it here and merely logging it we will no longer have any idea that there is a problem, and the user is in some weird state.

app/utils/user.js
```diff
@@ -47,6 +41,22 @@ const getExpirationTimeStamp = token => {
   return getCurrentUnixTimestamp() + exp - iat;
 };
 
+/**
+ * @returns string | void
+ */
+const refreshTokenIfExpired = async () => {
+  if (await tokenIsExpired()) {
+    const prev_refresh_token = await fetchItem('refresh_token');
+    const response = await postRequest('gesdinet_jwt_refresh_token', {
```
crucialfelix: This code is copied from the original, just flattened a bit.  It seems the problem is here then. That's what was raising all those errors.

We need to check the status of response and then the function needs to do the right thing.

400 ? So the user is not authorized?  That means Bad Request.  What was bad about the request? What is `prev_refresh_token` at this point for this Bad User?

Or was this a kind of 500 error (server bug) that returns a 400?

Regardless, the server response should be detected here and should probably return a null token. try catch is not the right way to write business logic, and that's kind of why this code is causing confusion and errors.

crucialfelix:
So what was the 400 error all about? Can I get logs from the server?

crucialfelix:
And how did you manage to trigger this error? I tested it with a legit login, then I edited the code to make the token expired so it would refresh and get a new token.

norbertschuler:
I am sorry, I currently do not know how I managed to get the 400 bad request error. I will have to examine this to find out how to reproduce it. Of course you're right, it's not a good solution to have the try/catch inserted - it's just otherwise it won't work and shows a blank page. If you could rewrite it to catch the error where the `postRequest` failed the way it does not stop code execution so the user can be shown the login page again, that would be great. Feel free to revert my commit with the try/catch block again.

crucialfelix:
Will do. btw. I find it easier to just get code review comments (which are always helpful). Adding commits confuses the development flow and adds maintenance time. Thanks

norbertschuler:
I think I know how to reproduce the 400 request. I edited my local storage in Chrome by changes the value of `token_expires` to 0 and also the value of `refresh_token` to 0. Then the following request returned the result below:

Request:
`POST https://www.trilliontreecampaign.org/api/token/refresh?version=v1.1&_locale=de`
with payload 
`{"refresh_token":"0"}`

Result:
```json
{"code":400,"message":"Validation Failed","errors":{"errors":["Bad credentials"],"children":{"_username":{"errors":["Bad credentials"]},"_password":[]}}}
```

crucialfelix:
I did the same kind of test, but I guess my token was still active. Yours was actually dead. 400 may be valid. Jorge will confirm.

I'll check response in refresh and then it's legit to return null. (not an error)

crucialfelix:
Why is it returning `{"_username":{"errors":["Bad credentials"]},"_password":[]}` for refresh token? That seems like the wrong response.

norbertschuler:
> Why is it returning `{"_username":{"errors":["Bad credentials"]},"_password":[]}` for refresh token? That seems like the wrong response.

That's a question I cannot answer as I have build the API. Probably something to optimize.

norbertschuler:
@crucialfelix Can I do anything to help you get this ready to merge?

crucialfelix:
Should be ready now. I moved the try catch to directly around the api call.

norbertschuler:
O.k., now if the `token_expires` value triggers a refresh and the `refresh_token` is invalid, the call to refresh the token `gesdinet_jwt_refresh_token` fails, but as long as the `token` is still valid, I am still logged in (before I was requested to login again). But I think that's o.k. - you just have errors caught in the browser until you logout and login again to refresh all this local variables.
The only thing bugging me is the error #1239 which occurred as I first checked out this branch and opened by Chrome browser. But I missed to investigate it and could not reproduce it now any more.
So let's merge this.

crucialfelix:
Re #1239 
You got a "Error · Request failed with status code 401" which is the correct response from the server: not authenticated.

All calls got getRequest or getAuthenticatedRequest get checked:
https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/2faff94a34801ef31958a2d4e96e70ad183f2175/app/utils/api.js#L21-L33

So I think that is correct behavior.

PR #1179, opened 2019-08-01
Fix redirection

niteshgrg:
Build for PR.
https://i.diawi.com/qrBNE5

### DIFF
--- a/app/components/Menu/index.native.js
+++ b/app/components/Menu/index.native.js
@@ -10,7 +10,7 @@ import {
 import { LargeMenuItem } from './MenuItem.native';
 import PropTypes, { func } from 'prop-types';
 import styles from '../../styles/menu.native';
-import { updateRoute, updateStaticRoute } from '../../helpers/routerHelper';
+import { updateRoute } from '../../helpers/routerHelper';
 import * as icons from '../../assets';
 import i18n from '../../locales/i18n.js';
 import { getLocalRoute } from '../../actions/apiRouting';
@@ -147,10 +147,7 @@ export default class Menu extends Component {
       );
     } else if (urlBreak[urlBreak.length - 2] === 'donate-trees') {
       this.props.selectPlantProjectAction(urlBreak[urlBreak.length - 1]);
-      setTimeout(
-        () => updateStaticRoute('app_donate_detail', navigation, 0),
-        0
-      );
+      setTimeout(() => updateRoute('app_selectProject', navigation, 0, {}), 0);
     } else {
       setTimeout(
         () => updateRoute('/' + urlBreak[urlBreak.length - 1], navigation, 0),
--- a/app/components/Menu/index.native.js
+++ b/app/components/Menu/index.native.js
@@ -10,7 +10,7 @@ import {
 import { LargeMenuItem } from './MenuItem.native';
 import PropTypes, { func } from 'prop-types';
 import styles from '../../styles/menu.native';
-import { updateRoute } from '../../helpers/routerHelper';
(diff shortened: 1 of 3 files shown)


norbertschuler:
Clicking the url https://www.trilliontreecampaign.org/donate-trees/1 on my iOS simulator I am directly redirected to the page to enter the amounts of trees to donate:
Simulator Screen Shot - iPhone X - 2019-08-01 at 22 28 22

@sagararyal Is that the desired behavior? I rather thought one should first see the project page as following:
Simulator Screen Shot - iPhone X - 2019-08-01 at 22 32 54

jmiridis:
The parameter (1) in `https://www.trilliontreecampaign.org/donate-trees/1` is intended to be used a the id of the supported treecounter, not as a plant project. The request to be able to handle 'supported treecounter' arouse first and so I provided a the route
`@Route("/donate-trees/{treecounterId}", name="app_donateTrees_support")`.
This route is transparent for the backend, meaning that it simply loads the application and it is the applications decision to interpret and handle the parameter.

When the necessity arouse for a route with a plan project parameter, I rejected adding another route and instead favored the solution of providing a route like
`@Route("/donate-trees/c/{context}", name="app_donateTrees_context")`

{context} will be some token or encrypted parameters that are passed to the application, and the application will have to make an API call to get the parameters behind the token. This way multiple parameters can be handled like: supported treecounter, plant project, minimum number of trees to donate, etc.

This is a work in progress and currently not available.

rarawal:
@norbertschuler ^

norbertschuler:
@jmiridis as https://www.trilliontreecampaign.org/donate-trees/1 is now already used in all the advertisement banner @sagararyal has created I don't think we can change it so quickly any more even if it is wrong. So we should fix the behavior of the app and client to use this way first.

norbertschuler:
If you rather want the behavior as the web client I would propose this change:
```patch
index 1c3f3940..5f154886 100644
--- a/app/components/Menu/index.native.js
+++ b/app/components/Menu/index.native.js
@@ -10,7 +10,7 @@ import {
 import { LargeMenuItem } from './MenuItem.native';
 import PropTypes, { func } from 'prop-types';
 import styles from '../../styles/menu.native';
-import { updateRoute, updateStaticRoute } from '../../helpers/routerHelper';
+import { updateRoute } from '../../helpers/routerHelper';
 import * as icons from '../../assets';
 import i18n from '../../locales/i18n.js';
 import { getLocalRoute } from '../../actions/apiRouting';
@@ -148,7 +148,7 @@ export default class Menu extends Component {
     } else if (urlBreak[urlBreak.length - 2] === 'donate-trees') {
       this.props.selectPlantProjectAction(urlBreak[urlBreak.length - 1]);
       setTimeout(
-        () => updateStaticRoute('app_donate_detail', navigation, 0),
+        () => updateRoute('app_selectProject', navigation, 0, {}),
         0
       );
     } else {

```

niteshgrg:
@jmiridis even i noticed above and was surprised by it but we already have this working in web app. I think the best decision for us to do is change the variable name in url to be more suitable. And not change the functionality.

sagararyal:
https://www.trilliontreecampaign.org/donate-trees/{id} or {slug}
should lead to the Project Page 

The donor can read about the project and make a donation.

@jmiridis  I propose /donate-trees/t/{id} to support a tree-counter, or decide this later with context.

sagararyal:
From: Jorgo
I am ok with it, knowing now that it has not been used for ‘supported treecounter’.
@niteshgrg @norbertschuler 
can be merged.

But /1 should take to project.

norbertschuler:
I‘ll wait for Nitesh to do the change.

PR #1181, opened 2019-08-01
put texts for tooltips

yllipetrovci:


norbertschuler:
I had to change the tooltip `id` of the `ReactTooltip` element to individual names as otherwise you can not use different tooltips on the same page. It only shows one text for both tooltips if you hover on them.

Issue #1188, opened 2019-08-02
Hide Captured CO2 when data = null


sagararyal:
### Info can't be clicked; please use the icons,

– Component is not responsive in web.

Please refer to designs
https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/872

<img width="592" alt="Screen Shot 2019-08-02 at 1 12 47 PM" src="https://user-images.githubusercontent.com/5401575/62366484-4d7e8500-b527-11e9-9f16-8053f5f65d6f.png">

Issue #1186, opened 2019-08-02
NDVI optimizations left from #1139


norbertschuler:
- In the iOS and Android app "October, 2015" is not translated on first view - but when you click a data point it gets replaced with correctly translated month names.
- Values for Min, Avg and Max should be rounded mathematically, e.g. two digits behind the dot (0,65544566 -> 0.66) - otherwise the layout looks very ugly.

![Screenshot_1564740970](https://user-images.githubusercontent.com/1532418/62363335-56804e00-b549-11e9-95a7-471599416809.png)

norbertschuler:
@yllipetrovci looks as if you already fixed that in your branch

Issue #872, opened 2019-06-04
NDVI and Biomass Components


sagararyal:
Build re-useable NDVI and Biomass Components (cards) for Web and Apps

to display NDVI time series
current NDVI data
Captured CO2

<img width="259" alt="Screen Shot 2019-06-04 at 10 36 40 AM" src="https://user-images.githubusercontent.com/5401575/58864268-ab702700-86b4-11e9-8a41-e4482d513723.png">

The components should be added to Plant Projects Page, and Contributions Page that match specific conditions eg: has `ndvi=available`

Also refresh ndvi values and page data on scroll

**NDVI Scale:**
For exact colors, please refer to the ui with color Scale: https://zpl.io/ankj77n

**Index**
-0.2 or less = water
0.1 or less= area of barren rock,sand, snow
0.2 to 0.5 = shrubs , grassland and senescence crops
0.6 to 0.9=dense vegetation

PR #1190, opened 2019-08-02
Bugfix/#1139

hkurra:
Fixed NDVI issues and disabled show user contribution page on if the contribution has NDVI data we will change it once we have contribution page ready


sagararyal:
I'd still recommend to revert the NDVI changes; and merge this to NDVI branch.

norbertschuler:
Fixed from #1186
- translation of month "October" on initial view in native app
- rounded values for max, avg, min in native apps

Not fixed from #1186
- no rounded values for max, avg, min in for web client

Fixed from #1188
- Round ndvi to 2 digits -> compare above
- Info can't be clicked; please use the icons
- Component is missing tile design in web client

Not fixed from #1188 
- Component is missing tile design in native app
Simulator Screen Shot - iPhone X - 2019-08-03 at 11 23 44
- Hide Captured CO2 card when data = null in native apps - for the web client I just got this crash #1193
Simulator Screen Shot - iPhone X - 2019-08-03 at 11 31 00
- Component is not responsive in web
Bildschirmfoto 2019-08-03 um 11 22 07

Compare to initial design in issue #872

Use the projects from WeForest and Plant-for-the-Planet on production for testing this!

norbertschuler:
Fixed:
- responsive layout in web client
- rounded values in web client

Not fixed:
- showing "null Kg" if data is missing in app and crashing in the web client #1200
- no tile design in native app as in web client (is that intended?)

PR #1197, opened 2019-08-04
Fix #1182 trying to replace the loading indicator with a SVG animation

### LINKED ISSUE #1182: Optimize Loading gif size
https://staging.trilliontreecampaign.org/c33d4dd2234fb1721a4ac0baa3bfb3c0.gif

Currently this gif which we use for loading screens uses 424.5KB 

We should to reduce this to less than 10-20kb, or find alternate loading animations.

norbertschuler:
This is just work in progress as the animation does not work in the native apps. Probably it's not supported in SVG files and needs to be done outside the SVG manually in a react component as in the native SvgContainer.
(compare #1182)


norbertschuler:
I have not found a solution for an animated SVG in the native apps. Closing this. At least I added a GIF animation half the size of the former one.

PR #1199, opened 2019-08-05
Feature: dotenv configuration

crucialfelix:
Configuration variables are loaded from `defaults.env`. You may add a local `.env` file to override these.

In React source code you can use these as: `process.env.BUGSNAG_API_KEY` which will then be replaced by their values at build time by a webpack plugin. For react-native (which does not use webpack) these are replaced by "babel-plugin-transform-inline-environment-variables"

`defaults.env` is in source control and should always work for any developer without further config.

`.env` can be created for any secure settings or for building releases.

This keeps API keys and other secure settings out of source control.

Only activates bugsnag (bug reporting) if `BUGSNAG_API_KEY` is set

Tested for web and react-native

This replaces: https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1081


crucialfelix:
Other things I discovered and tested:

webpack may be added for react-native using https://github.com/callstack/haul which seems to have good support. It's more of a change then we would want to do right now.

pro: would also enable image and asset optimization, typescript
con: more webpack config madness

`react-native-config` doesn't work with normal `react`
`env-cmd` I used for a bit but the `./with-env` script is easier and allows supports of defaults.

`with-env` could be modified to allow an environment name.  eg. `./with-env production react-native bundle ...`

crucialfelix:
ISSUE:
https://github.com/Plant-for-the-Planet-org/treecounter-app/wiki/Release-documentation

The build process requires using Android Studio / XCode. 

GRRR.  I didn't realize that until now.  Maybe there is some way to customize the build command to pre-pend either `./with-env` or `source "production.env"`

I have seen many people using build services who do 12 factor env var settings and generate the env file automatically as part of their CI/CD.

norbertschuler:
Yes, we don’t have automatic builds of the native apps and can’t build different environments through build tools for Android or iOS. If we find a solution capable of doing it all, I’d like to switch. But we’d need to find a way to introduce this without disturbing developers from building the features still on the roadmap.

crucialfelix:
The cure is now officially worse than the disease. React native + React web are a PITA sometimes. 

I will close this and replace with a simple PR that just enables `bugsnag` to be conditionally activated. We can go stay with config file.

crucialfelix:
Closing in favor of the simpler https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1228

Issue #880, opened 2019-06-05
Translation: localized number formatting


norbertschuler:
All numeric values should be displayed localized, e.g.

English: 0.2 Trees
German: 0,2 Bäume

English: 12,200,302 Trees
German: 12.200.302 Bäume

The same for amount of currency. There are libs for that purpose.

mahmed0715:
PR created.

mahmed0715:
updated pr as a feature branch. @norbertschuler

norbertschuler:
Localization of the web client was successfully finished. Now only more some localization for the native apps is missing, e.g. in these components I want to add here from Slack as a small reminder:

<img width="403" alt="Bildschirmfoto 2019-08-20 um 15 34 42" src="https://user-images.githubusercontent.com/1532418/63424147-1a386300-c40e-11e9-8967-df8901532c10.png">

mahmed0715:
I am working on Native part of this task.

norbertschuler:
@mahmed0715 can you please use my work in progress for this page (I tried to make the styles much simpler as they have been now):

https://github.com/Plant-for-the-Planet-org/treecounter-app/tree/feature/880_number_format

![Simulator Screen Shot - iPhone X - 2019-08-21 at 15 57 31](https://user-images.githubusercontent.com/1532418/63438762-2aac0600-c42d-11e9-8f88-a4fc4f94bf00.png)
![Simulator Screen Shot - iPhone X - 2019-08-21 at 15 58 34](https://user-images.githubusercontent.com/1532418/63438763-2aac0600-c42d-11e9-8578-d46a433dbc2e.png)

PR #1207, opened 2019-08-09
Fix number format and currency localization

mahmed0715:
#880 Updated, we formatted numbers based on the locale, and currency based on the locale + currency.
Checked donation page for currency.
@sagararyal  @norbertschuler Please check and advice

### DIFF
--- a/app/components/Common/NumberFormat.js
+++ b/app/components/Common/NumberFormat.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { currentUserProfileSelector } from '../../selectors';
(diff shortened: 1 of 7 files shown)


norbertschuler:
Please use GitFlow and name this branch "feature/localization-number-880".

mahmed0715:
Used gitflow and as the ticket's label as 'bug' I made the branch as bugfix/.
Should I go and change as feature branch?  @norbertschuler

norbertschuler:
Prove me wrong, but GitFlow doesn’t have the type bugfix as a branch type. That’s why I thought also fixes are named feature branches.

PR #1208, opened 2019-08-09
Fix number format and currency localization 

mahmed0715:
#880 Updated, we formatted numbers based on the locale, and currency based on the locale + currency.
Checked donation page for currency.
@sagararyal @norbertschuler Please check and advice 
New PR is created with feature branch.


mahmed0715:
Please ignore older PR with bugfix branch. Its newer with feature branch.

norbertschuler:
The longer German wording "Milliarde" unfortunately breaks the TreeCounter on the home page. As this is quite prominent, I think we must optimize this before we can release this patch:

Bildschirmfoto 2019-08-12 um 12 04 31
Bildschirmfoto 2019-08-12 um 12 05 50

norbertschuler:
On the page choosing the number of trees to donate (after you choose a project) I do not know why I cannot see a comma for the German translation (instead of a dot as in English). Maybe that's because the layout is cut or this value isn't formatted. The underlining and the small font of the "150" trees option is very strange anyway, but that's another issue:

Bildschirmfoto 2019-08-12 um 12 12 48
Bildschirmfoto 2019-08-12 um 12 13 11

norbertschuler:
If you continue the process, there comes another page where the currency and amount if not yet adapted:

Bildschirmfoto 2019-08-12 um 12 20 26

norbertschuler:
The native apps page to choose the amount of trees to donate seems to be separated from the web page and has not yet any localization done:

Simulator Screen Shot - iPhone X - 2019-08-12 at 12 24 44

app/locales/de/commonlabels.json
```diff
@@ -19,5 +19,11 @@
   "continue_on_site": "Weiter auf der Website",
   "error": "Fehler",
   "success": "Erfolg",
-  "congrats": "Glückwunsch"
+  "congrats": "Glückwunsch",
+  "Thousand": "Tausend",
+  "Million": "Million",
+  "Billion": "Milliarde",
+  "Trillion": "Billion",
+  "Quadrillion": "Quadrillion",
```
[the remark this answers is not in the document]
mahmed0715: which one should be Billiarde? I got all those from internet

mahmed0715:
Any suggestion on the long words issue with "Milliarde"? We can minimize the text size if that is "Milliarde"? or how @norbertschuler

norbertschuler:
> We can minimize the text size if that is "Milliarde"? or how @norbertschuler

Changing text size would be my favorite solution. I think the German wording is already quite long, but the longest translation I found in Spanish as "Mil millones". If you want to can adapt the text size to this, so future translation don't get the problem again.

norbertschuler:
As I recognized this PR left out the native app and concentrated on the web client. I would like to always make changes on both. If you agree please also adapt the native code (mostly same name as the web client with `.native.js` as file names. So the tree counter in the native apps still are not changed yet - @sagararyal what do you think of this topic?

Simulator Screen Shot - iPhone X - 2019-08-13 at 08 56 44
Simulator Screen Shot - iPhone X - 2019-08-13 at 08 56 11

app/locales/de/commonlabels.json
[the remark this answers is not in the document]
crucialfelix: Billiarde = 100 crores

app/utils/utils.js
```diff
@@ -7,3 +13,19 @@ export function delimitNumbers(str) {
...
+export function formatNumber(data, locale, currency) {
+  locale = locale || getLocale();
+  try {
+    let style = { maximumFractionDigits: 2 };
+    if (currency) {
+      style.style = 'currency';
+      style.currency = currency;
+    }
+    // console.log('got numberformat', data, locale, currency, style)
+    return new Intl.NumberFormat(locale, style).format(data);
```
crucialfelix: This works in ES6 and will work in browsers: https://caniuse.com/#search=Intl

I don't think React native will work on Android. It has an old shitty js
https://github.com/facebook/react-native/issues/15382

As that guy reports: if you run-android in the debugger then it will work, because the code is actually running in Chrome. But once you ship, it doesn't work on the device.

There is a polyfill here: https://github.com/andyearnshaw/Intl.js/

Maybe create `app/utils/numberFormat.js` and just `export Intl.NumberFormat`.
Then create `app/utils/numberFormat.android.js` which imports the polyfill and exports it.
Then import just that into utils.js

There is also: https://www.npmjs.com/package/react-native-intl
Provides a Native Intl implementation, but also includes translation. So it's too fat.

app/locales/de/commonlabels.json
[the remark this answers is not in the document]
mahmed0715: Okay we are doing this, Please advice
 "Thousand": "Tausend",
  "Million": "Million",
  "Billion": "Milliarde",
  "Trillion": "Billion",
  "Quadrillion": "Billiarde",
  "Quintillion": "Trillion"

app/utils/utils.js
mahmed0715: Thanks @crucialfelix Yes will do that, 
I think in this PR we do for web client only, then I will take the native part.
 okay? @sagararyal @norbertschuler

crucialfelix: It should be really simple, so it's much easier if you can just do it in this PR. We don't want to merge any PR that breaks react native.  

I can help you on chat if you are unsure what I meant above.

mahmed0715: Thanks, Yeah I will do that, if needed surely will get back to you, Thanks...

norbertschuler: If you can do the changes for the native apps also in this branch that would of course be awesome.

mahmed0715:
> On the page choosing the number of trees to donate (after you choose a project) the value for the "150" trees option is still formatted with a dot, not with a comma. The underlining and the small font of the "150" trees option is very strange anyway, but that's another issue:
> 
> Bildschirmfoto 2019-08-12 um 12 12 48
> 
> Bildschirmfoto 2019-08-12 um 12 13 11

@norbertschuler the small 150 text because its an input element, its text size is different(thats differnt issue), my explanation on that right side input with period(.) not comma(,) for German is that because we could not calculate a number with comma, I am making this to use comma but before calculation/next use of that value we must convert back it to period and store in database.. please advice if we really want to do that.  Please check https://stackoverflow.com/questions/35315157/html5-input-box-with-type-number-does-not-accept-comma-in-chrome-browser 
 
Big Question: Actually why do we need an input in right side as this is just to show the total cost calculated from the left input number of trees? we can make it like a simple text field to show just the amount in user's language.

norbertschuler:
> Big Question: Actually why do we need an input in right side as this is just to show the total cost calculated from the left input number of trees? we can make it like a simple text to show just the amount in user's language.

From my tests, the user can also enter the amount he wants to donate and then the number of trees is calculated from that number.

As currently most of the traffic (and probably the donations) is coming from users from Germany, a good localization is of course an issue. Before the whole project was only in English, but now as we have German translations the user also expects  an input behavior matching the language. This could be especially important when entering amount of money to pay as otherwise support issues could occur. If we can not enter the amount with German formatting I would indeed disable the possibility of entering it at all and only allow editing the tree number and change the amount field to a pure display field as you suggested.

app/utils/utils.js
mahmed0715: Yeah I am on it,  I am a beginner in React native but I will do that in this PR..
Thanks @all for your co-operation

mahmed0715:
@norbertschuler I think now its good with input font size and localization with it.
You please check your reported issues, thus web client should be good..
 I will try to implement this in react native..
Thanks

norbertschuler:
> You please check your reported issues, thus web client should be good..

You fixed all my issues for the web client. Also I would prefer not having a disabled input field in the third row of the donation details page for showing the amount (a user might wonder, why he can try to edit it but can not change it) but rather just a normal text displayed. But for now this quick fix also will do it. And perhaps @sagararyal says entering a free donation amount is essential for the users, so we have to rethink this again.

Looking forward to your change of the native apps code (I am even more beginner of React programming as you are.)

app/utils/utils.js
crucialfelix: If you want me to add the one commit to make it work for rn, let me know. If you want to learn a bit about rn, then this is a good time to do it. But you probably should install everything, and that will take a lot of time.

- Create `app/utils/Intl.js` with just `export default Intl`; (something like that)
- Create `app/utils/Intl.android.js` which imports the polyfill and exports that as default.
- Then you import that wherever you need to use `Intl`

sagararyal:
@norbertschuler @mahmed0715 It is important that users are able to choose a free number of trees they want to donate. (don't need to worry about the amount).

mahmed0715:
Yeah thats already persists and as @norbertschuler requested to format the input field content localized, its already done. so we kept the both side input field. @sagararyal

app/utils/utils.js
norbertschuler: I would like to approve this PR, but I am not sure if this issue is resolved now?

crucialfelix: Not yet. @mahmed0715 If you want, I can just add the one commit with the native-react support.

mahmed0715: sure, Thanks, That would be very helpful. Please do.

mahmed0715:
> @norbertschuler @mahmed0715 It is important that users are able to choose a free number of trees they want to donate. (don't need to worry about the amount/currency).

We allowed users to do both input tree/amount.

app/utils/utils.js
crucialfelix: Done, just need to test a bit. Will push the commit tomorrow.

norbertschuler:
> Yeah thats already persists and as @norbertschuler requested to format the input field content localized, its already done. so we kept the both side input field. @sagararyal

I tested it  again and both input field are editable. The field for the donation amount (not the amount of trees) does just not allow to input float numbers but just integer numbers. So the user does not have to care about entering a dot or a comma. But on the other hand if the price is like 0.57€ and you enter 150 trees, a donation amount of 85,5€ is display in the input field for the donation amount. If you try to edit this, you can only delete and overwrite it with integer number again. You are not able to modify the floating number. But I think that's o.k. for now.

app/utils/utils.js
norbertschuler: Great, I'd like to test once you pushed it.

crucialfelix:
I've added the polyfill for react-native. It is actually only need for android, but I have it activated for both ios and android so that it gets tested by more people.

But now I realize that this function isn't used anywhere on native yet.

eg.
https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/4e46e608f13277e71f6036675101e1280b06a3e6/app/components/PlantProjects/PlantProjectSnippet.native.js#L192-L199

it's just displaying the raw number.

Maybe we should merge this now and then do another feature for the native app changes.

mahmed0715:
@crucialfelix thanks, Yeah,  as the polyfill added, I will be able to use the INTL implementation in rn app next feature. I will take care of this next time.

app/actions/getLocale.native.js
```diff
@@ -1,9 +1,25 @@
...
+/**
+ * Call this when the app starts up
+ * so that locale is cached. Most important
+ * on react-native where getting languages is slower.
+ */
+export async function initLocale() {
+  cache.locale = await guessLocale();
+}
+
+export function getLocale() {
```
crucialfelix: This was previously async. That was a problem because getLocale is caused many times on each render (and there are still a lot of duplicate renders to fix).

Now it's cached: initLocale() is called once when the app starts

app/components/PlantProjects/PlantProjectSnippet.native.js
```diff
@@ -191,10 +175,10 @@ class PlantProjectSnippet extends React.Component {
 
               <View style={styles.costContainer}>
                 <Text style={styles.costText}>
-                  {this.currency_symbols[currency]
-                    ? this.currency_symbols[currency]
-                    : currency}{' '}
-                  {specsProps.treeCost}
+                  <NumberFormat
```
crucialfelix: This correctly formats prices with currency symbols, falling back to the code (eg. KES) if unknown.

crucialfelix:
It now works on react-native everywhere I could see. Turned out to be quite easy.

A separate issue is if we want the currencies displayed in the user's donation currency. Probably yes. That's another issue.

A component like this would convert for display in user's preferred currency.
```<UserPrice data={tree.price} currency={tree.currency} /> ```

norbertschuler:
> It now works on react-native everywhere I could see.

No, there are still native pages missing without number formatting. Easiest way to spot them is to check where @mahmed0715 made changes in the web client and check if there are *.native.js file with the same name not adapted.

PR #1228, opened 2019-08-12
Conditionally activate bugsnag only if api key is in config

crucialfelix:
This gets the bugsnag api key from `config/index.js` and activates bugsnag reporting only if it's set.

Avoids all developers sending bugsnag errors by default from their local machines.

This is a simple alternative to the problematic attempt to use 12 factor env vars: https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1199 ( which seemed like a good idea at the time ).

### DIFF
--- a/app/components/ErrorBoundry/globalErrorBoundry.js
+++ b/app/components/ErrorBoundry/globalErrorBoundry.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { info } from '../../assets';
+import { context } from '../../config';
 import i18n from '../../locales/i18n.js';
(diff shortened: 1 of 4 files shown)


crucialfelix:
btw. I keep calling it bugsense because I used to use a different service called that.

crucialfelix:
Yes, this just fixes what it says: bugsnag defaults to off.

We do need a better way to switch config files for different build targets. That's another issue.

norbertschuler:
I just changed `index.js` for all four backends now and merged develop into staging triggering a rebuild. Then I will purge https://staging.trilliontreecampaign.org/ at Cloudflare and check if Bugsnag is still reporting.

PR #1229, opened 2019-08-12
Fixed #947 Crash due to wrong Video Id and don't show youtube player …

### LINKED ISSUE #947: com.facebook.react.bridge.JSApplicationIllegalArgumentException in MainActivity

**com.facebook.react.bridge.JSApplicationIllegalArgumentException** in **MainActivity**
Error while updating property 'videoId' of a view managed by: ReactYouTube

[View on Bugsnag](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5d0c9c889142b600208d88af?event_id=5d0c9c8800445ff6b5090000&i=gh&m=ci)

## Stacktrace

    ViewManagersPropertyCache.java:95 - com.facebook.react.uimanager.ViewManagersPropertyCache$PropSetter.updateViewProp

[View full stacktrace](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5d0c9c889142b600208d88af?event_id=5d0c9c8800445ff6b5090000&i=gh&m=ci)

hkurra:
…on web if video Id is not valid

ALso fixed re-rendering issue due to video on Android
project causing issue : "Königswald Ontananga"

### DIFF
--- a/app/components/Common/VideoContainer.native.js
+++ b/app/components/Common/VideoContainer.native.js
@@ -43,6 +43,9 @@ class VideoContainer extends React.Component {
     // native module that is very sensitive to being covered or even touching its containing view.
     setTimeout(() => {
       this.setState({ moduleMargin: StyleSheet.hairlineWidth });
+    }, 2000);
+    setTimeout(() => {
+      this.setState({ moduleMargin: StyleSheet.hairlineWidth });
     }, 250);
     if (this.props.onReady) this.props.onReady(event.nativeEvent);
   };
--- a/app/components/Common/VideoContainer.native.js
+++ b/app/components/Common/VideoContainer.native.js
@@ -42,7 +42,7 @@ class VideoContainer extends React.Component {
     // controls. We also use the minimal margin to avoid `UNAUTHORIZED_OVERLAY` error from the
     // native module that is very sensitive to being covered or even touching its containing view.
     setTimeout(() => {
-      this.setState({ moduleMargin: StyleSheet.hairlineWidth });
+      this.setState({ moduleMargin: StyleSheet.hairlineWidth + 1 });
     }, 2000);
     setTimeout(() => {
       this.setState({ moduleMargin: StyleSheet.hairlineWidth });
--- a/app/components/Common/VideoContainer.native.js
+++ b/app/components/Common/VideoContainer.native.js
@@ -9,6 +9,7 @@ class VideoContainer extends React.Component {
     this.state = { moduleMargin: StyleSheet.hairlineWidth * 2 };
   }
   componentWillMount() {
+    let videoId = undefined;
     if (this.props.url) {
       let ID = '';
       let url;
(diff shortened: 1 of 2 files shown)


norbertschuler:
Please name your branches `feature/<name>` using GitFlow.

app/components/Common/VideoContainer.native.js
```diff
@@ -52,7 +55,7 @@ class VideoContainer extends React.Component {
             this._youTubeRef = component;
           }}
           style={[{ height: 300 }, { margin: this.state.moduleMargin }]}
-          onYouTubeReady={this._onReady}
+          onReady={this._onReady}
```
norbertschuler: This change had the effect that on Android the play button and other controls are not shown any more when then video is displayed!

hkurra: Let me verify, There is not any supported Props name "onYouTubeReady"

norbertschuler: Weird behavior of my Samsung Galaxy S6 Edge - has anyone another device to test this?

norbertschuler: I added some logging into the code and on my S6 the code in the method `_onReady` does not get executed if the prop  `onYouTubeReady` is used (as you say it does not exists, so it is probably just ignored). But not executing the code of `_onReady` make my device show the controls. If the code is executed on the other hand using the correct prop `onReady` I always get the error of the missing YouTube controls. 

Maybe the developers of the YouTube component already fixed the bug the code of the method `_onReady` was trying to fix (see the long comment!) and isn't necessary any more (at least for Android)?

(On my Android emulator both versions of the code just work fine.)

app/components/Common/VideoContainer.native.js
```diff
@@ -18,31 +19,36 @@ class VideoContainer extends React.Component {
...
-  _onReady(event) {
+  _onReady = event => {
     clearInterval(this._interval);
 
     // The Android YouTube native module is pretty problematic when it comes to mounting correctly
     // and rendering inside React-Native's views hierarchy. For now we must trigger some layout
     // changes to force a real render on it so it will smoothly appear after ready and show
     // controls. We also use the minimal margin to avoid `UNAUTHORIZED_OVERLAY` error from the
     // native module that is very sensitive to being covered or even touching its containing view.
+    setTimeout(() => {
```
norbertschuler: Sorry, this additional lines do not fix the problem for me.

hkurra: But this is working fine for me if u understood correctly it as working fine earlier because on ready was not getting called so our timer was getting called after every 250 ms which was causing some rendering issue in whole Pag, I have fixed that bug by providing correct method, the ultimate goal here is to Re-render after some time to render the control

hkurra: can you make this change to verify if it is working or not in your device
image

norbertschuler: It's indeed working if I add your latest change `+1`.

norbertschuler: Can you explain what this code actually does? I do not understand why two times setTimeout is used now? Also which device needs this fix as every device I tested also worked without this the method `_onReady`?

hkurra: See this video for the behavior on my emulators after changes I suggested 
https://drive.google.com/file/d/18NRmyyhpyJpooKJYZ1pibHyERi2cRXBj/view?usp=sharing

hkurra: ° `Why I made this change`: earlier _onready was not getting called and the interval was not getting clear and donate page dancing, I think  someone reported this issue earlier
° No, after I made changes we are again re-rendering after clearing timeout but with only 250 ms and I think Video taking more than 250 ms to be in a ready state so now I have added 2 timeouts first with 250 ms as recommended by library another one with 2000 ms to wait for some more time so that Youtube player/video is ready to receive update
° this might be a hack but this is the best solution I found till now, else we can revert the change and let them dance

norbertschuler: O.k., I understood. Funny finding: Now my S6 seems to already work without the additional `setTimeout` (I tested by deleting the lines with the new `setTimeout` again - probably loading the video for a second time is faster as some caches accelerating the initial loading time). But let's keep it now as there will be a lot of old phones with longer loading times. Can you commit the change with the `+1` before merging this?

norbertschuler: Perhaps you also can add this explanation as a comment for other developers.

hkurra: Sure, will do that

PR #1233, opened 2019-08-14
added CircleCI config to build branches without backend

Plant-for-the-Planet:
I added a configuration for building all GitFlow branches (for code validation) which do not yet have a backend. Both the source for the web client as the bundles for the iOS and Android are build to detect errors.

These builds do not use any resources of our production server, while the protected branches `master`, `staging`, `test` and `devel` are currently directly being build on our production server and therefore **use up CPU power from the live system**.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,10 +93,29 @@ jobs:
           name: Install npm
           command: npm install
       - run:
-          name: Build Application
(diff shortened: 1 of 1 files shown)


Plant-for-the-Planet:
I assume errors introduced in the code for the native apps would not be recognized as this is only building the web client with webpack.

PR #1237, opened 2019-08-15
Web performance optimizations

crucialfelix:
## 1. SvgContainer (the globe of trees with animated pots): Removed unnecessary `forceUpdate();`

It looks like that was left over from a previous version that animated by changing state and forcing a rendering.
Since 1 year ago it toggles classes directly on the DOM (which is much faster), so forceUpdate() is unnecessary.

length of intro animation:

before: 19 seconds (14s scripting)
after: 6 seconds (365ms scripting) which is the natural length

Web feels much lighter now on that page.
Native is a different implementation.

## 2. `DonateTrees` dog-piling requests to `fetchCurrencies`

Both DonateTrees and SelectPlantProject would call fetchCurrencies on componentDidMount. This meant there were two simultaneous network requests, they would both return and cause a full page render. Most of the time it isn't even needed because it's already in the redux store.

- Call only if currencies is null.
- Throttle fetchCurrencies so only the first request sends. If the request does fail (server down) then currencies is still null and 10 seconds later it can be tried again as the user navigates elsewhere. On donate-trees if currencies is null it just doesn't do currency conversion and doesn't sort them by EUR.

This already makes the app feel much lighter.

## 3. Memoize `trillionCampaign`

Similar issue to 2 - this only needs to be called once per visit / app session. It was refetching every time you go back and forth between pages which makes everything feel very slow.

Maybe we could add a `cachedGetAuthenticatedRequest` for these API calls. I would do that in a future optimization. We need a function that fetches only if needed, only once, caches with a reasonable timeout (6 hours or something). Currently as you navigate around the app it's fetching stuff over and over.

## 4. Add why-did-you-update to track unnecessary React updates

You can activate this in `index.web.js` by uncommenting the code. Posts notifications in console about possibly avoidable renders. It's a bit noisy, so its off by default.  I will add this and some other notes to documentation later.

## 5. fix: "list should have a unique key prop" in GiftTrees and Trillion

Gets rid of big red errors in console. Helps React avoid unnecessary renders because it doesn't know what order the children element were in.

## 6. SelectPlantProject: refactor componentWillReceiveProps and componentWillMount to a single method

There is actually a better way to structure this to reduce re-renders of this large component on every key stroke but I will work on that in a refactor DonateTrees PR.

There are many other places I've found for optimizations, especially on native. 

## TO TEST THIS

Look at each page:

- Trillion Trees front page (Globe trees animation)
- Donate Trees
- Gift Trees

Look for errors in console.

### DIFF
--- a/app/components/PlantProjects/PlantProjectSnippet.native.js
+++ b/app/components/PlantProjects/PlantProjectSnippet.native.js
@@ -139,7 +139,10 @@ class PlantProjectSnippet extends React.Component {
             countTarget={specsProps.countTarget}
           />
           <View style={styles.projectSpecsContainer}>
-            <View key="a" style={styles.projectNameContainer}>
+            <View
+              key="projectNameContainer"
+              style={styles.projectNameContainer}
+            >
               <Text
                 ellipsizeMode="tail"
                 numberOfLines={1}
@@ -159,7 +162,10 @@ class PlantProjectSnippet extends React.Component {
                 />
               ) : null}
             </View>
-            <View key="b" style={styles.projectdetailsContainer}>
+            <View
+              key="projectdetailsContainer"
+              style={styles.projectdetailsContainer}
+            >
               <View style={styles.locationContainer}>
                 <Text style={styles.locationText} ellipsizeMode="tail">
                   {specsProps.location}
@@ -199,8 +205,8 @@ class PlantProjectSnippet extends React.Component {
               </View>
             </View>
 
-            <View key="c" style={styles.actionContainer}>
-              <View key="a" style={styles.byOrgContainer}>
+            <View key="actionContainer" style={styles.actionContainer}>
+              <View key="byOrgContainer" style={styles.byOrgContainer}>
                 <Text
                   style={styles.byOrgText}
                   ellipsizeMode="tail"
@@ -211,7 +217,7 @@ class PlantProjectSnippet extends React.Component {
               </View>
 
               {this.props.plantProject.allowDonations ? (
-                <View key="b" style={styles.buttonContainer}>
+                <View key="buttonContainer" style={styles.buttonContainer}>
                   <PrimaryButton
                     style={styles.buttonItem}
                     buttonStyle={styles.buttonStyle}
--- a/app/components/PlantProjects/PlantProjectSnippet.native.js
+++ b/app/components/PlantProjects/PlantProjectSnippet.native.js
@@ -139,7 +139,7 @@ class PlantProjectSnippet extends React.Component {
             countTarget={specsProps.countTarget}
           />
           <View style={styles.projectSpecsContainer}>
-            <View style={styles.projectNameContainer}>
+            <View key="a" style={styles.projectNameContainer}>
               <Text
                 ellipsizeMode="tail"
                 numberOfLines={1}
@@ -159,7 +159,7 @@ class PlantProjectSnippet extends React.Component {
                 />
               ) : null}
             </View>
-            <View style={styles.projectdetailsContainer}>
+            <View key="b" style={styles.projectdetailsContainer}>
               <View style={styles.locationContainer}>
                 <Text style={styles.locationText} ellipsizeMode="tail">
                   {specsProps.location}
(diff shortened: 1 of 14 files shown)


crucialfelix:
This shows part of the issue with fetchCurrencies. The first time it renders, the projects are not sorted by price. Then it gets currencies, then it re-renders (in the blink of an eye) with the new sort order.

Screen Shot 2019-08-14 at 1 28 49 PM

Screen Shot 2019-08-14 at 1 31 46 PM

crucialfelix:
Trillion trees SVG animation:

### before
before

### after
after

The red markers mean that browser was blocking and the frame rate drops.

sagararyal:
The web app does feel much faster, but at the same time, I saw the bundle.js is getting bigger and bigger.

Screen Shot 2019-08-15 at 12 31 46 PM

sagararyal:
I believe one of the reasons `forceUpdate()` was done is because when you switch from one tree counter to another, the animation state didn't really change.
case:
**going from:** 
Screen Shot 2019-08-15 at 12 41 51 PM


**to**
Screen Shot 2019-08-15 at 12 42 03 PM

in the second the green bar should not be dark green.

I think forceupdate() allowed the state to be:

Screen Shot 2019-08-15 at 12 43 39 PM


I could be wrong, but @harshkurra @niteshgrg can share more.

crucialfelix:
Sagar was talking about the feature to search for a user and see the goal animated:
https://www.trilliontreecampaign.org/t/sagar-aryal

it's the same on this branch as it is on production.  

There are other issues with flickering that I can look into.

norbertschuler:
> The web app does feel much faster, but at the same time, I saw the bundle.js is getting bigger and bigger.

This 7.26MB file is the not optimized bundle size you have on a development system - which server is this screen shot showing? The bundle size on production is now 2.9MB (compressed to 716KB thanks to Cloudflare) after our latest release 1.0.5 (formerly about 4MB and compressed 1.1MB). For the staging, test and devel servers I provide the sources map directly on the server, so the browser can just download them directly for debugging while on the production system they are not available but only uploaded into Bugsnag.

Bildschirmfoto 2019-08-15 um 13 26 29

crucialfelix:
Also this PR doesn't really increase bundle size.  We will definitely get code splitting and tree shaking working and reduce bundle size soon.

hkurra:
@crucialfelix @norbertschuler let me review these changes once

app/components/PlantProjects/PlantProjectSnippet.native.js
```diff
@@ -199,8 +199,8 @@ class PlantProjectSnippet extends React.Component {
               </View>
             </View>
 
-            <View style={styles.actionContainer}>
-              <View style={styles.byOrgContainer}>
+            <View key="c" style={styles.actionContainer}>
```
hkurra: it's better if we use a style name as a key to have better readability and to avoid the same key in the future development like here we could have used actionContainer

app/components/TreecounterGraphics/Trillion.native.js
```diff
@@ -207,8 +212,9 @@ class Trillion extends PureComponent {
 
   render() {
     return [
-      this.props.navigation ? (
+      this.props.navigation && (
```
hkurra: @crucialfelix @norbertschuler we have used ternary operator deliberately here, we have faced compilation issue on android due to this earlier, that may be due to the old version of react  but this is something I remember

app/components/PlantProjects/PlantProjectSnippet.native.js
crucialfelix: Good idea. Done

app/components/TreecounterGraphics/Trillion.native.js
crucialfelix: Maybe it is this ? https://github.com/facebook/react-native/issues/20764

I'm not sure if that would even apply here, but I changed it back to be safe.

I think the issue might have been:
```
        {'' && <Text style={styles.paragraph}>I am here</Text>}
```

React does conditional rendering safely if the value is a boolean, null or undefined. 
0 or "" are falsey but React has to render them (everybody would freak out if it doesn't render 0 or ""), so it does not short circuit. And on Android that means it needs to be wrapped in a <Text></Text>

hkurra: @crucialfelix I don't remember the exact issue because it was someone else who faced these issues may be @niteshgrg or @shubhi15, what I remember we faced some compilation issues due to this

PR #1246, opened 2019-08-15
use CircleCI resources to build the client 

norbertschuler:
Currently we use our production server to build the client. 

This patch makes CircleCI do the work and only copy the results to our server.


norbertschuler:
This would make https://github.com/Plant-for-the-Planet-org/treecounter-platform/blob/develop/.circleci/integrate-app.sh unnecessary and all the ~/source/ttc-app* folder on the server could be deleted.

PR #1258, opened 2019-08-17
App banner

mahmed0715:
@sagararyal app banner added for ios and android for web client.
Google play store app is linked up but still need the apple app id.
Please check and advice..


mahmed0715:
app id linked ios/google play/Translation added, @norbertschuler @sagararyal  please test and advice.

sagararyal:
@mahmed0715 

Screen Shot 2019-08-17 at 9 02 04 AM
1. is it possible to have the text and image link to App /play Store as well

2. I could not replicate it working on iOS devices user agents. (did try your updated code).

3. for how long will the banner disappear once the user closes. can we bring it back in a new session. / new window?

4. Some minor details on the button, align-> right
  
And maybe do not make the title bold `Plant-for-the-Planet App`
Screen Shot 2019-08-17 at 9 13 13 AM
add some padding top/ bottom maybe 1 em or 16 px?

5. Can we float it at the bottom of the screen? – It wont hide the login/signup button + is reachable by fingers in most tall phones.

norbertschuler:
Looks like this branch shares a commit 92393711fb220ef559f87fbb12d4753fafdf8ee2 with your other pull requests #1208. Was the number formatting necessary for the app banner to work? Hopefully this won't result into a merge conflict later on. I would prefer every feature to be based on the default branch and do not depend on each other.

mahmed0715:
@sagararyal all the requsted changes are pushed, I double checked in ios simulator, it should work now.
the displayDelay is 1 day(so after the close button is clicked it will show up next day), do we really need to show in new session?
other wise all other changes are good. we customized the component.
Please check and advice

web/stylesheet/index.scss
```diff
@@ -44,6 +44,10 @@
 
 @import "~slick-carousel/slick/slick.css";
 @import "~slick-carousel/slick/slick-theme.css";
+@import "~react-smartbanner/dist/main.css";
+// this will overwrite the default banner style
+@import "./app_banner.scss";
```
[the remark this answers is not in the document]
crucialfelix: no hound, it shouldn't

web/stylesheet/_variables.scss
```diff
@@ -12,3 +12,4 @@ $colorRedeemBorder: #9fc356;
 $colorRedeemInside: #f5fbe8;
 $boxShadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 3px 1px -2px rgba(0, 0, 0, 0.2),
   0 1px 5px 0 rgba(0, 0, 0, 0.12);
+$avocado: #89b53a;
```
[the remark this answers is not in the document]
crucialfelix: true, always end with a trailing newline.

web/stylesheet/_app_banner.scss
```diff
@@ -0,0 +1,49 @@
...
+    text-shadow: none;
+    font-weight: 700;
+    line-height: 1rem;
+    font-size: 0.9rem;
+    letter-spacing: 1px;
+  }
+  .smartbanner-author, .smartbanner-description {
+    @extend .pftp-text-basic;
+    color: $bannerAuthorColor;
+    text-shadow: none;
```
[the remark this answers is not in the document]
crucialfelix: Not important.

web/stylesheet/_app_banner.scss
```diff
@@ -0,0 +1,49 @@
...
+$bannerCloseButtonColor: #707070;
+$bannerAuthorColor: #202125;
+.smartbanner-container {
+  background: $bannerBackground;
+  text-shadow: none;
+  .smartbanner-close {
+    background: none;
+    border: none;
+    box-shadow: none;
+    color: $bannerCloseButtonColor;
```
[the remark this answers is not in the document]
crucialfelix: css and scss conventions do say to use banner-close-button-color. But keep to whatever the style app is currently using

app/components/SmartBanner/SmartBanner.js
```diff
@@ -0,0 +1,394 @@
...
+      windows: 'msApplication-ID',
+      kindle: 'kindle-fire-app'
+    }
+  };
+
+  constructor(props) {
+    super(props);
+
+    // if (!__SERVER__) {
+    ua = require('ua-parser-js'); // eslint-disable-line global-require
```
crucialfelix: Why are you requiring these in the constructor rather than importing at the top?  Did you see this recommended somewhere?

There may be problems with webpack and on demand loading of this code. If you wanted to delay loading (for efficiency) then it has to be done a different way (with import('ua-parser-js') and probably using Suspense and React.lazy())

app/components/SmartBanner/SmartBanner.js
```diff
@@ -0,0 +1,394 @@
...
+  componentWillMount() {
+    this.setType(this.props.force);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.force !== this.props.force) {
+      this.setType(nextProps.force);
+    }
+    if (nextProps.position === 'top') {
+      window.document
```
crucialfelix: assign this once to a variable.
```
let html = window.document.querySelector('html');
html.classList.add ...
html.classList.remove...
```

or better yet add a method to your class that will add / remove classes on `html`. You do this many times in the class, so it will be worth it to have a single method to do that.

crucialfelix: Also you only need to call these if this.props.position !== nextProps.position

mahmed0715: Actually @crucialfelix this file is from npm module react-smartbanner and to enable click we did a small customisation on it, I think we can move it into a new folder like /app/customised/SmartBanner ?

mahmed0715: please advice if we should do the requested changes.

mahmed0715:
new PR created https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1260 with git fixed. @norbertschuler @all
We can discuss there I think.

app/components/SmartBanner/SmartBanner.js
crucialfelix: oh I see, you just copied it over.  well then just leave it. I didn't mean to sound picky :)  But the require in side the constructor seems wrong.

crucialfelix: They did it only because they do server side rendering SSR:
https://github.com/patw0929/react-smartbanner/blob/729dfe4c478ae98c25345d1d9dc69a9fcf77d2af/src/components/SmartBanner.js#L83-L86

so it needed to be turned off when it's running on the server. 

For us those two should just be imported at the top of the file. It may work fine where it is, but it seems like it will bite us someday in the future.

crucialfelix: Don't worry about moving the file. But please put a big comment block at the top explaining that it's from the package but with a small change

and link to github and include the license: https://github.com/patw0929/react-smartbanner/blob/master/LICENSE as a big block of text. I think that should be correct and polite.

norbertschuler: To add the license of a new library we use, we even have a process described here (at the end): https://github.com/Plant-for-the-Planet-org/treecounter-app/wiki/Release-documentation - but the license menu is only displayed in the native apps, not in the web client yet.

mahmed0715: Added blocks in the file. adding licence information and the cause to add the file in code base.

sagararyal:
@mahmed0715 should we close this, now we have a different PR for the same.

mahmed0715:
Yes @sagararyal we can close this one.

Issue #1174, opened 2019-07-31
Implement Smart App Banners


sagararyal:
# Implement for both iOS & Android
![smartbanner_2x](https://user-images.githubusercontent.com/5401575/62206708-b08ae300-b392-11e9-8fd8-c76a11a2ad17.png)

https://github.com/patw0929/react-smartbanner

```
title: 'Plant-for-the-Planet App',
          author: 'Plant-for-the-Planet',
          button: 'VIEW',
          store: {
              ios: 'On the App Store',
              android: 'In Google Play',
          },
          price: {
              ios: 'FREE',
              android: 'FREE',
```

[Documentation from Apple](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/PromotingAppswithAppBanners/PromotingAppswithAppBanners.html)

sagararyal:
Design available at:
https://zpl.io/25AmwZQ

Should bee responsive, the button on the right,
and rest float on left.

PR #1260, opened 2019-08-18
New App banner implementation: fixed git issue

mahmed0715:
Here is the new PR for App Banner #1174


app/components/SmartBanner/index.js
```diff
@@ -0,0 +1,393 @@
...
+      android: 'google-play-app',
+      windows: 'msApplication-ID',
+      kindle: 'kindle-fire-app'
+    }
+  };
+
+  constructor(props) {
+    super(props);
+
+    // if (!__SERVER__) {
```
crucialfelix: The original package wanted to be friendly for SSR. We don't need or want that.

Move ua and cookie to normal imports up at the top of the file.

app/components/SmartBanner/index.js
```diff
@@ -0,0 +1,393 @@
+import React, { Component } from 'react';
```
crucialfelix: Add a big block comment here saying it's copied from react-smartbanner in order to do a small modification.

Add license information at the top as well: https://github.com/patw0929/react-smartbanner/blob/master/LICENSE

Actually you *could* inherit from it and just overwrite the methods you wanted to change the click behavior.

app/config/index.js.dist
```diff
@@ -1,22 +0,0 @@
-/**
```
norbertschuler: You accidentally removed app/config/index.js.dist from your branch. Please restore it as it was, so the build won’t fail any more.

mahmed0715: added again. Thanks..

web/index.html
```diff
@@ -27,10 +28,20 @@
     //   });
     // });
   </script>
+  <meta name="apple-itunes-app" content="app-id=1444740626">
+  <!-- <meta name="apple-itunes-app" content="app-id=284882215"> -->
+  <meta name="google-play-app" content="app-id=org.pftp">
+  <meta name="msApplication-ID" content="82a23635-5bd9-df11-a844-00237de2db9e">
+  <meta name="msApplication-PackageFamilyName" content="facebook_9wzdncrfhv5g">
+  <meta name="kindle-fire-app" content="app-id=B0094BB4TW">
```
norbertschuler: @sagararyal Do we offer the app for Windows Phone and Amazon? I didn't know that before. Where do these IDs (`82a23635-5bd9-df11-a844-00237de2db9e` for Windows and `app-id=B0094BB4TW` for Amazon) come from?

norbertschuler: Ups, actually the IDs are from the Facebook App, e.g. https://www.amazon.de/Facebook/dp/B0094BB4TW and http://www.windowsphone.com/s?appid=82a23635-5bd9-df11-a844-00237de2db9e - I would rather leave these values empty in stead of redirecting our users visiting the website on an Amazon and Windows device to Facebook.

mahmed0715: Removed those unnecessary app refs. @norbertschuler

sagararyal:
@mahmed0715 works fine for me, as expected. I think we can push to develop If @norbertschuler  or @crucialfelix  agree.

norbertschuler:
I am afraid I have some issues:

1. The text "Plant-for-the-Planet App" and the German button text "Erhalten" got cut on my iPhone X Simulator (also when testing the layout with the Chrome Dev Tools you can use to test smaller screen sizes):
iPhone X Simulator

2. Clicking the "Get"-Button works in my Chrome browser, but as I clicked in in my iPhone X Simulator, I got an error as following (in English: "Safari could not open the page as it is invalid"):
iPhone X Simulator 2

3. Compiling and running the native app with this branch results into the following error message on my iPhone X Simulator:
iPhone X Simulator 3

norbertschuler:
The cut of the button also happens if you make your Safari browser on MacOS pretend to be an iPhone:
Bildschirmfoto 2019-08-20 um 12 11 28

sagararyal:
@norbertschuler 
doesn't show up in actual safari on Mac OS.
Private Window

Screen Shot 2019-08-20 at 12 56 48 PM

sagararyal:
@norbertschuler 
doesn't show up in actual safari on Mac OS.
Private Window

Screen Shot 2019-08-20 at 12 56 48 PM

sagararyal:
On Safari 13.0
  ## pretending iphone
Screen Shot 2019-08-20 at 1 01 50 PM

## iPad
Screen Shot 2019-08-20 at 1 02 32 PM

norbertschuler:
@mahmed0715 for the latest issue (3) I now pushed a commit which added a separate `index.native.js` file for the native apps which I think do not need to promote themselves with a smart banner and also this prevents the crash (which might also could have been solved somehow else).

app/locales/de/commonlabels.json
```diff
@@ -19,5 +19,8 @@
   "continue_on_site": "Weiter auf der Website",
   "error": "Fehler",
   "success": "Erfolg",
-  "congrats": "Glückwunsch"
+  "congrats": "Glückwunsch",
+  "Plant_for_the_planet_App": "Plant-for-the-planet App",
```
norbertschuler: up until now we only used lower case names with _ between single words

app/components/SmartBanner/index.js
norbertschuler: I am sorry, but I am thinking it is very problematic to copy this code from the node module while still importing the module by`"react-smartbanner": "^5.1.3",` in package.json and using other parts of it directly from the imported package e.g. `@import "~react-smartbanner/dist/main.css";` in `web/stylesheet/index.scss`.

If later on some other developers decides to upgrade this package and do not see that you copied some code manually, the code might break. So either import the module and modify the class by inheritance (as @crucialfelix suggested) or copy everything in our project and don't import the module at all. Or as a third option, fork the public repository of https://github.com/patw0929/react-smartbanner and make your own modified version and import that into our project.

mahmed0715:
@norbertschuler Yeah I am going to update and publish a package and use it in our project.

for the app url in ios , i just cheked with locale en/de
https://apps.apple.com/us/app/id1444740626
https://apps.apple.com/de/app/id1444740626

and in browser it looks okay. I am adding a console.log with url for test purpose so that we know whats going on on your devices.

norbertschuler:
Regarding my (2) issue I just tested the URL https://apps.apple.com/us/app/plant-for-the-planet/id1444740626?l=de&ls=1 successfully on my real iPhone X where it works well while the same URL returns the error message on the iOS simulator. So I think this is just a simulator issue and works on real phones. The node module still uses the old iTunes URLs (`https://itunes.apple.com/${this.props.appStoreLanguage}/app/id `) but should also work well on real devices. This would be an example where I would expect this module to get an update soon which we would miss if we just copy the code.

norbertschuler:
Regarding my (1) issue here is a proposal for another banner design (which of course also did a change to the Smartbanner class directly):
Bildschirmfoto 2019-08-20 um 14 57 42

Commit https://github.com/Plant-for-the-Planet-org/treecounter-app/commit/2aed35c574c08b03d5cc25bfd236ad5b2e9ed78b by https://github.com/Plant-for-the-Planet-org/treecounter-app/tree/feature/1174-fixed_norbert

mahmed0715:
@norbertschuler I think I have fixed the button on German, the css is fixed I hope. Let me update PR,

mahmed0715:
@norbertschuler @sagararyal  please check if your issues are resolved

norbertschuler:
All my issues are resolved except that on the very big screen of an iPhone X (not talking about iPhone 5 or 6) the banner text is cut:

Bildschirmfoto 2019-08-20 um 17 24 12

If you don't have another idea about this (reduce font size for example), I would change the German translation to "Los!" (which is "Go!") or you maybe think again about using a icon or emoji here. But having the text cut on such big phones is not good for me.

mahmed0715:
Yeah I think changing German translation would be better in this situation. @norbertschuler

mahmed0715:
Please Check @norbertschuler if your issue resolved now, I changed german translation to use GO(LOS)

norbertschuler:
As I said, every kind of text currently is cut - may decrease font size? @sagararyal what do you think?

Bildschirmfoto 2019-08-21 um 09 23 01

mahmed0715:
@norbertschuler it should be good now, I decresed font size a little bit (from .9rem to .85 rem) and the width is adjusted. 
Please check...

PR #1271, opened 2019-08-20
Feature/add stripe react element

niteshgrg:


norbertschuler:
@yllipetrovci @niteshgrg can you please add a description and I think you haven't merged the latest .hound.yml from the develop branch and that's why you got again all these wrong ES6 review messages. Maybe you can merge the latest develop branch into this pull request?

norbertschuler:
Please merge develop into your branch to get rid of Hound *wuff*

app/server/routes/fos_js_routes.json
```diff
@@ -1,4399 +1,4885 @@
 {
-  "base_url": "",
```
norbertschuler: Have you actually changed anything in this file. Due to reformatting it's difficult to see.

web/config/webpack.common.config.js
```diff
@@ -42,36 +42,36 @@ const config = {
...
-      filename: './donatetreewidget.html',
-      template: './web/widget.index.html'
-    }),
-    new HtmlWebPackPlugin({
-      inject: true,
-      chunks: ['ndviwidget'],
-      filename: './ndviwidget.html',
-      template: './web/widget.index.html'
-    }),
+    // new HtmlWebPackPlugin({
```
norbertschuler: I assume you commented this to accelerate the build or do you really wanted to delete our widgets with this PR?

app/actions/apiRouting.js
```diff
@@ -12,7 +12,7 @@ export const getApiRoute = async (routeName, params) => {
   params =
     'api_login_check' === routeName
       ? params
-      : { version: 'v1.1', _locale: locale, ...params };
+      : { version: 'v1.3', _locale: locale, ...params };
```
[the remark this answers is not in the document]
niteshgrg: So as per my understanding nothing else should have changed in v1.3. But lets go through the app once to see if anything else do break in 1.3 if we can have this deployed on some environment it would be easier to get this whole app tested out. @norbertschuler

web/config/webpack.common.config.js
niteshgrg: Yes it was in draft i have uncommented it now.

app/actions/apiRouting.js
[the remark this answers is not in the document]
norbertschuler: Last time we did this for a PR for pledge events, I think we got in trouble at least for the menu API call and I decided only to increment the version for the single pledge event API call. Now only this call in our develop branch currently uses v1.3 while every other call is using v1.1 - also keep in mind this is not the only position where the version is defined. There are two exceptions in competition and challenge explicitly setting the API version to v1.1 when they got called.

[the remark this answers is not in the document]
niteshgrg: Ok i will check then. For now will change for just the api's i use.

app/components/StripePayment/CCForm.js
```diff
@@ -0,0 +1,92 @@
...
+      <input
+        type="checkbox"
+        className="payment-save-later-checkbox"
+        name="save-for-later"
+        checked={props.saveForLater}
+        onChange={() => {
+          props.onClickSaveForLater('saveForLaterCC');
+        }}
+      />
+      Save for later
```
norbertschuler: Here a translation resource should be used and some styling and spacing would be nice. What is the purpose if this checkbox (I am unsure if the user understand it)?

Bildschirmfoto 2019-09-12 um 09 37 17

app/components/Payment/PaymentSelector.js
```diff
@@ -57,196 +54,273 @@ class PaymentSelector extends React.Component<{}, { elementFontSize: string }> {
...
-            {!giftToName && (
-              <div>
-                {i18n.t('label.donate_to', {
-                  name: gatewayProps.context.plantProjectName
-                })}
+            );
+          }
+          if ('stripe_sepa' === gateway) {
+            return null;
+            // return (
```
norbertschuler: Will this return later on as you commented it for now?

app/components/Payment/PaymentSelector.js
```diff
@@ -57,196 +54,273 @@ class PaymentSelector extends React.Component<{}, { elementFontSize: string }> {
...
-                  />
-                </div>
-              );
-            }
-          })}
-        </div>
-      </StripeProvider>
+            );
+          }
+          if ('offline' === gateway) {
```
norbertschuler: Is there any project allowing us to test this or isn't it used any more?

app/components/StripePayment/CheckoutForm.js
```diff
@@ -0,0 +1,219 @@
...
+            style={{ arrow, displayNone, fontSize: this.props.fontSize }}
+            onClickSaveForLater={this.onClickSaveForLater}
+            saveForLater={state.saveForLaterCC}
+            cards={this.state.cards}
+            chosenCard={this.state.chosenCard}
+            currentUserProfile={this.props.currentUserProfile}
+            onChangeSelectedCard={this.onChangeSelectedCard}
+          />
+        ) : (
+          <SEPAForm
```
norbertschuler: Just to be sure once more: I think you commented SEPA in the code at https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1271/files#r324127744 - so do we support it or don't we support it?

app/components/Payment/PaymentSelector.js
norbertschuler: Sagar said not supported any more by now.

app/components/StripePayment/CheckoutForm.js
norbertschuler: Sagar said not supported any more by now.

app/components/Payment/PaymentSelector.js
norbertschuler: not used yet according to Sagar

PR #1276, opened 2019-08-21
Fixed all important tree counter Widget issue, after this anybody can test widget and log more defects

hkurra:
- Fixed tree counter Widget issue it was not loading the CSS(somehow CSS import was commented) 

- Fixed User profile circular image issue 

- Fixed issue where if project id is null then widget was making a donation page path with /null at the end of URL

How to test?

- Open http://localhost:8080/treecounterwidget.html to verify the changes 
- Actual testing can only be done after merging into staging

### DIFF
--- a/web/widgets/Treecounter/app/components/App.js
+++ b/web/widgets/Treecounter/app/components/App.js
@@ -108,7 +108,7 @@ export default class App extends Component {
         <style>{`:host {all: initial;}`}</style>
         <div className="widget-container" id={'widget-container'}>
           <link href="treecounterwidget.css" rel="stylesheet" />
-          {/* <link href={`${serverName}/treecounterwidget.css`} rel="stylesheet" /> */}
+          <link href={`${serverName}/treecounterwidget.css`} rel="stylesheet" />
           {/* Apply CSS hooks here */}
           <style>{style}</style>
           {/* Apply React Tooltip Library CSS */}
@@ -136,7 +136,7 @@ export default class App extends Component {
                 onClick={event => {
                   const url = `${serverName}${getLocalRoute(
                     'app_donateTrees'
-                  )}/${projectId}`;
+                  )}${projectId ? '/' + projectId : ''}`;
                   window.open(url, '_blank');
                 }}
               >
--- a/web/widget.index.html
+++ b/web/widget.index.html
@@ -1,15 +1,20 @@
 <!DOCTYPE html>
 <html>
-
-<head>
-  <meta charset="utf-8" />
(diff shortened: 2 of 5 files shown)


web/widgets/Treecounter/app/components/App.js
```diff
@@ -108,7 +108,7 @@ export default class App extends Component {
         <style>{`:host {all: initial;}`}</style>
         <div className="widget-container" id={'widget-container'}>
           <link href="treecounterwidget.css" rel="stylesheet" />
-          {/* <link href={`${serverName}/treecounterwidget.css`} rel="stylesheet" /> */}
+          <link href={`${serverName}/treecounterwidget.css`} rel="stylesheet" />
```
crucialfelix: So this stylesheet is now linked twice, once locally and once on the server?  Also I'm trying to figure out what URL I can see this at. This is `App` but inside web/widgets and I don't yet know where that stuff is used.

hkurra: Why 2 links ? 
First one help us in local development and second one is for production, first one will not work in production or in another way you will see a error in network tab, second one is must to make it work in production 
Why it is inside App.js?
Widget is a separate Application which should in any environment without any further assistance, through this approach widget JS bundle will load the css by itself user just have to import the JS file.

See the documentation https://docs.google.com/document/d/1RDl5MZhz6HXgjU2lRTfUBu9tie-7iCqIxO22SvmEg0Y/edit?usp=sharing
or 
this CodePen to understand  it 
https://codepen.io/Hkurra/pen/BbqROx

crucialfelix: So one of them will always throw an Error. But that's not the end of the world.

```
Failed to load resource: the server responded with a status of 403 ()
widget.js:69 GET https://cdpn.io/boomboom/v2/widget.css net::ERR_ABORTED 403
(anonymous) @ widget.js:69
Promise.then (async)
(anonymous) @ widget.js:49
(anonymous) @ widget.js:83
```

hkurra: > So one of them will always get a network error. But that's not the end of the world.
> 
> ```
> Failed to load resource: the server responded with a status of 403 ()
> widget.js:69 GET https://cdpn.io/boomboom/v2/widget.css net::ERR_ABORTED 403
> (anonymous) @ widget.js:69
> Promise.then (async)
> (anonymous) @ widget.js:49
> (anonymous) @ widget.js:83
> ```

Yes, correct

Issue #448, opened 2018-11-14
Safari Bug with Register Tree Date


sagararyal:
on Safari, date field dropdown doesn't show up.

<img width="827" alt="screen shot 2018-11-14 at 9 57 22 pm" src="https://user-images.githubusercontent.com/5401575/48511955-74d9d880-e858-11e8-801d-c4172f07b9c0.png">

This might help, it is because safari and IE does not support html5 datepicker natively.

https://caniuse.com/#search=date

https://www.tjvantoll.com/2012/06/30/creating-a-native-html5-datepicker-with-a-fallback-to-jquery-ui/
http://www.javascriptkit.com/javatutors/createelementcheck2.shtml

sagararyal:
<img width="224" alt="Screen Shot 2019-07-13 at 2 04 25 PM" src="https://user-images.githubusercontent.com/5401575/61171395-375c5500-a577-11e9-9ab8-d5b36375021c.png">
A fallback for safari could be something like this.

PR #1277, opened 2019-08-21
Fixed bug #448 : Added react-datepicker library and create custom Tem…

hkurra:
…plate with custom CSS

See https://github.com/Hacker0x01/react-datepicker for more info


hkurra:
@crucialfelix @norbertschuler please provide your initial feedback on this

norbertschuler:
I think the backend can not deal with the format returned by the date picker (error says "the value should not be null"):

Bildschirmfoto 2019-08-22 um 16 30 50

norbertschuler:
Works for me on MacOS with Safari and Chrome :-)

app/components/Templates/DatePickerTemplate.js
```diff
@@ -1 +1,44 @@
-export { TextInputTemplate as DatePickerTemplate } from './TextInputTemplate';
+import React from 'react';
+import i18n from '../../locales/i18n';
+import { formatDate, getDateFromMySQL } from '../../helpers/utils';
```
crucialfelix: We could consider using `fecha` https://www.npmjs.com/package/fecha which is very small (and could replace the famously bloated `Moment`).

In any case I don't think we should be writing our own date parsers/formatters. I just had to examine all the code and double check everything. That's the cost of rolling your own.

`formatDate` here is Date => `YYYY-MM-DD` 
`getDateFromMySQL` is `YYYY-MM-DD` => Date

Both of those are juggling the 0-indexed JavaScript month.

moment has parse and format https://momentjs.com/docs/#/parsing/string-format/
https://momentjs.com/docs/#/displaying/

norbertschuler:
@hkurra please check our reviews by time if this pull request should ever get merged.

app/components/Templates/DatePickerTemplate.js
```diff
@@ -1 +1,42 @@
...
+import { getLocale } from '../../actions/getLocale';
+// import { compCalendar } from '../../assets';
+import DatePicker from 'react-datepicker';
+
```
norbertschuler: In order to show a localized German date picker we would have to add the following here:
```
import { registerLocale } from  'react-datepicker';
import de from 'date-fns/locale/de';
registerLocale('de', de)
```
We would have to add the npm module `date-fns` as `react-datepicker` depends on it instead of Moment.js - so we would have two! date formatting libraries making our bundle size again 3MB instead of 2,6MB:

Bildschirmfoto 2019-09-25 um 14 21 25

norbertschuler: If we wan't to continue this patch, I propose trying to exchange all usage of `Moment.js `with `date-dns`.

norbertschuler: Here are the hints you need to implement date-fns: https://date-fns.org/v2.4.1/docs/I18n and https://date-fns.org/v2.4.1/docs/webpack

app/components/Templates/DatePickerTemplate.js
norbertschuler: I now replace Moment.js with date-fns as the used DatePicker from react-datepicker required it anyway.

app/components/Templates/DatePickerTemplate.js
```diff
@@ -1 +1,44 @@
...
+import 'react-datepicker/dist/react-datepicker.css';
+
+export function DatePickerTemplate(locals) {
```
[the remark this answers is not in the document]
norbertschuler: Me neither, but at least it works, so I won't change it for now.

app/components/NDVI/NDVIfunctions/parseDate.js
```diff
@@ -1,59 +0,0 @@
-import moment from 'moment';
-import 'moment/min/locales';
-import { getLocale } from '../../../actions/getLocale';
-
-const parseDate = (month, year) => {
```
norbertschuler: This was really weird code: Why should we display "January 1 - 31 January" - that's just January and everyone should know how many days  January has. So I just simplified this to showing on the month and year anymore and omit the days.

app/utils/utils.js
```diff
@@ -46,6 +48,19 @@ export function formatNumber(data, locale, currency, userProfile, currencies) {
...
+
+  if (date) {
+    return format(parseISO(date), style, {
+      locale: localeObjects[locale]
+    });
+  } else {
+    return '';
+  }
+}
+
```
mahmed0715: We are sure we will always use ISO?

app/utils/utils.js
```diff
@@ -46,6 +48,19 @@ export function formatNumber(data, locale, currency, userProfile, currencies) {
...
+  locale = locale || getLocale();
+  console.log('formatDate', date, style, locale);
+
+  if (date) {
+    return format(parseISO(date), style, {
+      locale: localeObjects[locale]
+    });
+  } else {
+    return '';
+  }
```
mahmed0715: It would be better if this code be enclosed with try catch statement so that if something is wrong on the parsing or formatting, we are good to go without crashing..

app/server/parserLiformToTcomb.js
```diff
@@ -129,6 +130,11 @@ export default function parseJsonToTcomb(liformSchemaJson, config, validator) {
           case 'date':
             options.template = DatePickerTemplate;
             options.type = properties[propertyKey].widget;
+            options.config = {
+              ...options.config,
+              format: date => formatDate(date, 'dd MMM yyyy'),
+              dateFormat: date => formatDate(date, 'dd MMM yyyy')
+            };
```
mahmed0715: My proposal: Do we need to pass the format all the time? I think es6 will use default.

norbertschuler: You're right and I already have the same format 'dd MMM yyyy' as a default in the function definition. I am going to change that.

app/utils/utils.js
norbertschuler: The date values to parse are coming directly from the backend (no user input). I already checked for the case of an undefined value, but if the parsing fails, it would indicate a severe problem of the data returned by our API, so I'd like to get a crash caught by Bugsnag to be able to spot the problem as it would also be an error for our users out there.

app/components/NDVI/NDVIfunctions/parseDate.js
norbertschuler: @sagararyal was it an imported feature in the specification to always show the number of days of a month instead just showing the month, e.g. `1-31. Januar 2016` instead of `January 2016` or `1-30. June 2019` instead of `June 2019`?

PR #1278, opened 2019-08-21
Fix: getLocale() - initialize on demand

crucialfelix:
PR https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1208
assumed that locale could be initialized on app startup,
but i18n.js calls getLocale() before that. Regardless of what calls first,
this initializes locale if needed.

### DIFF
--- a/app/actions/getLocale.js
+++ b/app/actions/getLocale.js
@@ -11,6 +11,9 @@ export function initLocale() {
 
 export function getLocale() {
   // console.log('getLocale', cache.locale);
+  if (!cache.locale) {
+    initLocale();
+  }
   return cache.locale;
 }
 


crucialfelix:
I checked it. 

The native version doesn't call getLocale(), instead it does:
```
let userLang = undefined;
getLanguages().then(languages => {
  userLang = languages[0].split('-')[0];
  i18next.changeLanguage(userLang); // ['en-US', 'en']
})
```
 so it implements it's own async check of languages, guesses which one is first, then calls changeLanguage. This is also not ideal, but I I want to go slow with changes especially before a release.

norbertschuler:
> so it implements it's own async check of languages, guesses which one is first, then calls changeLanguage. This is also not ideal, but I I want to go slow with changes especially before a release.

That's why it was working *lol*

PR #1279, opened 2019-08-21
Fix #880 localization numbers in native apps

### LINKED ISSUE #880: Translation: localized number formatting

norbertschuler:
starting localizing amounts from #880 
- improved layout of donation tree amount chooser
- @mahmed0715 changed the formatting for payment, tree counter and more
- added some more number formatting for counters and removed formatting in input fields to allow to enter higher numbers as 9999

### DIFF
--- a/app/components/Challenge/Tabs/ChallengeUser.native.js
+++ b/app/components/Challenge/Tabs/ChallengeUser.native.js
@@ -2,7 +2,6 @@ import React, { Component } from 'react';
 import { ScrollView, View, TextInput, Text } from 'react-native';
 import SearchUser from './SearchUser.native';
 import PrimaryButton from '../../Common/Button/PrimaryButton';
-import { delimitNumbers } from '../../../utils/utils';
 import ChallengeList from '../challengeList';
 import CardLayout from '../../Common/Card';
 import { Dropdown } from 'react-native-material-dropdown';
@@ -144,6 +143,8 @@ class ChallengeUser extends Component {
               />
               <View style={challengeStyles.flexContainerStyle}>
                 <Text>{i18n.t('label.challenge_to_plant')} </Text>
+              </View>
+              <View style={challengeStyles.flexContainerStyle}>
                 <TextInput
                   keyboardType="numeric"
                   underlineColorAndroid={'transparent'}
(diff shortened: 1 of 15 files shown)


app/utils/utils.js
```diff
@@ -59,3 +60,36 @@ export function isIOS() {
...
+        ' ' + i18n.t('label.Trillion'),
+        ' ' + i18n.t('label.Quadrillion'),
+        ' ' + i18n.t('label.Quintillion')
+      ][x / 3]
+    );
+  } else if (x > 9) {
+    return delimitNumbers(n / 1000000000) + ' ' + i18n.t('label.Billion');
+  } else if (x > 6) {
+    return delimitNumbers(n / 1000000) + ' ' + i18n.t('label.Million');
+    // TODO: think about using the label "thousend" - under 1 Mio it's uncommon to use it
```
mahmed0715: Yes we did that I think,  I mean under 1 Million we use Thousand, and its applicable for all other like billion

norbertschuler: My comment was irritating. I mean in German you would use 1.6 Million but you rather would write 10,000 instead of 10 Thousand.

app/components/Challenge/Tabs/ChallengeUser.native.js
```diff
@@ -144,17 +143,19 @@ class ChallengeUser extends Component {
...
+              <View style={challengeStyles.flexContainerStyle}>
                 <TextInput
                   keyboardType="numeric"
                   underlineColorAndroid={'transparent'}
                   style={challengeStyles.treecount_input}
                   onChangeText={evt => {
                     this.handleTreeCountChange(evt);
                   }}
-                  value={delimitNumbers(this.state.treeCount)}
+                  value={String(this.state.treeCount)}
```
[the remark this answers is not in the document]
norbertschuler: Yes, we are not formatting the user input, so the user writes something like 100000 - otherwise I couldn’t enter such a high value. I find it ok to do so.

app/components/Currency/TreeCountSelector.js
```diff
@@ -160,7 +160,7 @@ class TreeCountSelector extends React.Component {
             <input
               type="text"
               disabled={this.state.isFixed}
-              value={delimitNumbers(this.state.variableAmount)}
+              value={String(this.state.variableAmount)}
```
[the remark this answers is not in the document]
norbertschuler: Same as above

app/utils/utils.js
norbertschuler: Better example: Would you prefer reading 2,75 Thousand or 2,750 ?
But this is only my personal feeling about what sounds better in the German language and not that important issue. That’s why I made a TODO of it.

PR #1280, opened 2019-08-21
Feature/880 native for Number format

mahmed0715:
Please check the native part, I tried to fix all the places I see, if you find anything please report.
@norbertschuler

### DIFF
--- a/app/components/PlantProjects/PlantProjectSnippet.native.js
+++ b/app/components/PlantProjects/PlantProjectSnippet.native.js
@@ -10,7 +10,8 @@ import { getImageUrl } from '../../actions/apiRouting';
 import { targetPlanted, tick, questionmark_orange } from '../../assets';
 import TouchableItem from '../Common/TouchableItem.native';
 import PlantedProgressBar from './PlantedProgressbar.native';
-import NumberFormat from '../Common/NumberFormat';
+import { formatNumber } from '../../utils/utils';
+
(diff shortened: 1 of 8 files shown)


app/components/PlantProjects/PlantProjectSnippet.native.js
```diff
@@ -181,10 +182,7 @@ class PlantProjectSnippet extends React.Component {
 
               <View style={styles.costContainer}>
                 <Text style={styles.costText}>
-                  <NumberFormat
-                    currency={currency}
-                    data={specsProps.treeCost}
-                  />
+                  {formatNumber(specsProps.treeCost, null, currency)}
```
norbertschuler: Why did you replace the `NumberFormat` component with a direct call to `formatNumber` here? I just want to understand when to use each of them correctly so I do it correctly too.

mahmed0715: Actually you can see in the NumberFormat component we call the util method formatNumber, so it doesnt matter if you use the component or utils. @norbertschuler 

Feedback requested

norbertschuler: Thank you. So I think we should use the component if suitable and not use the method `formatNumber` directly.

norbertschuler:
thank you, I will merge this into my branch

PR #1287, opened 2019-08-23
Feature: native performance optimizations

crucialfelix:
Optimizing performance and simplifying some components with react hooks.

Rewrote SelectPlantProject/Tabs
- featured.native.js
- list.native.js
- price.native.js
- country.native.js

This shows how much nicer hooks makes things. `useMemo` greatly simplifies and optimizes those.

---

Updates `react-redux` and related libraries to latest versions.  This adds `useDispatch` and `useSelector`.

I was looking for optimization wins by eliminating updates to the big containers and moving that redux access down to smaller function containers/components that display them.

This could allow things like:

```js

// Display Price in user's preferred currency and locale number formatting
export default Price = ({ price, currency }) => {
   // get stuff we need from redux.
   const { userCurrency, currencies, userLocale }  = useSelector(mapStateToProps);

   const converted = convertCurrency(price, currency, userCurrency, currencies);
   return NumberFormat(converted, userLocale, userCurrency);
};

const mapStateToProps = state => {
 const  userProfile =  currentUserProfileSelector(state);
 return {
  userCurrency: userProfile.currency,
  userLocale: userProfile.locale,
  currencies: state.currencies,
 };
};
```

Use it like this:
```js
<Price price={project.treeCost} currency={project.currency} />
```

Now whenever the currencies or the user's locale or preferred currencies are changed, all prices on the page update. It means you don't have to pass around locale, userProfile and currencies through all these deeply nested components.

I didn't use any redux hooks yet, but this makes it possible for future PRs.

locale and user currency aren't stored there yet. 
Using it for `currencies` would be a big win in cleaning up code.

I think https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1304 would benefit from this.

---

Add utils/currency.js : `convertCurrency` and `sortProjectsByPrice`

---

In Donate Trees ( SelectPlantProject.index.native.js ) there is a temporary solution for lazy loading the tabs. The issue here is that all tabs were loading and that is a lot of big list views with pictures.

react-native-tabs-views 2 supports lazy rendering and native swipe gestures.

I tried to update to that but it requires to update the pods. I will do that in a separate PR. There are several other places where tabs should be lazy rendered.

---


app/components/SelectPlantProject/Tabs/list.native.js
```diff
@@ -1,107 +1,80 @@
...
+            clearButtonMode="while-editing"
+            underlineColorAndroid={'transparent'}
+            onChangeText={setSearch}
+            value={search}
+            autoCapitalize="none"
+            autoCorrect={false}
+            returnKeyType="search"
+            placeholder={i18n.t('label.searchshort')}
+            placeholderTextColor={placeholderTextColor || '#ccc'}
+            style={[styles.searchInput, { paddingVertical: 0 }]}
```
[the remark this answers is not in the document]
hkurra: @crucialfelix I want to review this but not possible before tomorrow let me know if it is fine else merge it as per your understanding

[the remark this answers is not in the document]
crucialfelix: I'm still working on it for a bit more, no hurry. I just wanted to get it up so others could see it. I will request a review from you when it's ready. Thanks!

app/components/PlantProjects/PlantProjectDetails.native.js
```diff
@@ -1,74 +1,99 @@
...
 import { link } from '../../assets';
 import TouchableItem from '../../components/Common/TouchableItem';
-import i18n from '../../locales/i18n';
+import VideoContainer from '../../components/Common/VideoContainer';
 import NDVI from '../../containers/NDVI/NDVI';
+import i18n from '../../locales/i18n';
+import styles from '../../styles/selectplantproject/plant-details.native';
+import PlantProjectImageCarousel from './PlantProjectImageCarousel';
+
+const cleanUrl = url => {
```
crucialfelix: AFAICT links to TPO were not working at all. Most of them don't start with `https://` so the `Linking` said that it can't open them.

This function prepends https:// if it looks like a URL

app/components/SelectPlantProject/index.native.js
```diff
@@ -64,16 +48,20 @@ export default class SelectPlantTabView extends Component {
...
-  _renderSelectPlantScene = ({ route }) => {
+  renderSelectPlantScene = ({ route }) => {
+    const props = ({ plantProjects, onMoreClick, selectProject } = this.props);
+    const { index } = this.state;
+
+    // Only render a tab if it is focused
     switch (route.key) {
       case 'featured':
-        return <FeaturedProjects {...this.props} />;
+        return index === 0 && <FeaturedProjects {...props} />;
```
crucialfelix: This is the hack to only render a tab if it's focused.  Do give it a good test: go back and forth quickly.

I will submit a PR to update react-native-tab-view later.

app/containers/SelectPlantProject/index.js
```diff
@@ -1,78 +1,78 @@
...
+  //   );
+  //   let userPreviousDonations = nextProps.userSortedContributions.filter(
+  //     project => project.contributionType === 'donation'
+  //   );
+  //   if (userPreviousDonations.length > 0) {
+  //     let selectedProject = plantProjects.filter(
+  //       project => project.id === userPreviousDonations[0].plantProjectId
+  //     );
+  //     if (selectedProject.length > 0) {
+  //       //TODO hkurra commenting this to fix donation on APP
```
crucialfelix: @hkurra Do you remember what the issue was and what this does?  I left it commented out but not removed.

PR #1288, opened 2019-08-23
feature/plant project snippet design

harshvitra:


app/components/PlantProjects/PlantedProgressbar.native.js
```diff
@@ -62,34 +73,40 @@ class PlantedProgressBar extends React.Component {
...
+           ) : null} */}
           </View>
         </View>
-
         {!this.props.hideTargetImage ? (
           <View style={styles.targetContainer}>
             <Text style={styles.treePlantedtext}>
-              {countTarget ? delimitNumbers(countTarget) : null}
+              {countTarget
+                ? i18n.t(numeral(countTarget).format('0.0 a'))
```
crucialfelix: I think we needed locale dependent formatting here. It's mainly the Germans who write 1,6 instead of 1.5 like the rest of us ;)  

Should we not use the same number formatting everywhere? I already had to install a polyfill to get Intl.NumberFormat to work on Android. Is `numeral` better?

harshvitra: Hey as of now I was focusing on the design, the numeral might not work because of other issues like it gives only the letter m for million so it wont be translated.. 

This pull is mainly to change the design

app/components/PlantProjects/PlantProjectSnippet.native.js
```diff
@@ -10,6 +10,8 @@ import { getImageUrl } from '../../actions/apiRouting';
 import { targetPlanted, tick, questionmark_orange } from '../../assets';
 import TouchableItem from '../Common/TouchableItem.native';
 import PlantedProgressBar from './PlantedProgressbar.native';
+import NumberFormat from '../Common/NumberFormat';
```
norbertschuler: This import does not seem to be used, is it?

app/components/PlantProjects/PlantedProgressbar.native.js
norbertschuler: @harshvitra you are introducing a new npm module http://numeraljs.com/ with this patch. Is this necessary? Can't we format the number with the given libraries which @mahmed0715 and @crucialfelix introduced lately with our already merged _number localization pull request_?
I would like to keep the number of used modules as low as possible to avoid further increase of the already quite big bundle size of our web client.

harshvitra: Yes, I added this just to showcase design. Once design is approved I'll remove the code and the number localization can be looked that.

app/components/PlantProjects/PlantProjectSnippet.native.js
harshvitra: Will be removed

app/components/PlantProjects/PlantedProgressbar.native.js
norbertschuler: @harshvitra if you only want to create a pull request to present some code for testing next time you can create it as a "draft" which later on can be changed to a real pull request (but it does not work the other way).

app/components/PlantProjects/PlantProjectSnippet.native.js
crucialfelix: You should use the `NumberFormat` component rather than `delimitNumbers`.  

I'm investigating whether it will be better to get locale from redux directly inside NumberFormat. It will be much more efficient and easier to maintain.

norbertschuler:
Please rename his branch from `hotfix/plantProjectSnippetDesign` to `feature/plantProjectSnippetDesign` and make it a draft pull request.

Issue #1286, opened 2019-08-23
NaN on second page of treecounter for null values


norbertschuler:
**Describe the bug**
Instead of a zero the numbers on the second page of a tree counter are shown as "NaN".

**Is it on a Production or Development Build?**
staging (develop)

**To Reproduce**
1. Go to  a tree counter on staging with second page e.g. https://staging.trilliontreecampaign.org/t/norbert-schuler
2. Click on the > to navigate to the second page
3. See "NaN" values 

**Expected behavior**
Show number values, e.g. "0".

**What actually happens**
"NaN" is displayed.

**Screenshots**
![Bildschirmfoto 2019-08-23 um 09 56 09](https://user-images.githubusercontent.com/1532418/63576733-1ed75580-c58d-11e9-9aaf-a3d810867dd4.png)
![Bildschirmfoto 2019-08-23 um 09 55 53](https://user-images.githubusercontent.com/1532418/63576746-23037300-c58d-11e9-9907-2deb38d1d94a.png)

**Desktop (please complete the following information):**
- OS: MacOS latest
 - Browser: Chrome latest
 - Plant-for-the-Planet App Version 1.0.5

sagararyal:
in the tree counter you receive the following:

```
"countPlanted": 52025,
    "countPersonal": 40041,
    "countReceived": 11984,
```

In this case it should not be 0.

von mir / by me = `countPersonal`

Mir gewidmet = `countReceived`

mahmed0715:
I am on it.

norbertschuler:
Note to myself: "always test the code before merging" :-) Looks like there is still the need for some enhancements here ;-)

<img width="581" alt="Bildschirmfoto 2019-08-23 um 19 28 49" src="https://user-images.githubusercontent.com/1532418/63611507-486b9d80-c5dc-11e9-81c2-5e975fc31445.png">

norbertschuler:
I assume we can just decrease the font size as on for the numbers on the first page of the counter.

mahmed0715:
Yes this is what we do . If its greater than 12 chars we decrease size. Here may be we didnt check this. Will do asap.

PR #1292, opened 2019-08-23
Fix plantproject details NAN fix, removed unused delimitnumbers

mahmed0715:
Please check this quick fix.


mahmed0715:
For #1286

mahmed0715:
The thing is, we do delimitnumbers using utils method convertNumbers and pass this as props to ProjectDetails component, then again we did the delimitnumbers in there, that can nott parse the already delimitted numbers.
I think we can go ahead and merge. @norbertschuler

PR #1306, opened 2019-08-26
Some more fixes for tree counter numbers

norbertschuler:
- fixed "NaN" value for tree counter details for native clients
- unified null value checks for tree counter numbers
- adapted arrow check (enable link to detail page) for web client which was different from native apps, so the link was almost never shown

### DIFF
--- a/app/components/TreecounterGraphics/TreecounterGraphicsText.js
+++ b/app/components/TreecounterGraphics/TreecounterGraphicsText.js
@@ -65,7 +65,7 @@ class TreecounterGraphicsText extends Component {
                 {this.props.trillion ? (
                   <div>
                     {/* {this.getTwoWordString(NumberToWords.toWords(target))} */}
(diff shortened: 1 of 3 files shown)


app/components/TreecounterGraphics/TreecounterGraphicsText.js
```diff
@@ -87,13 +87,13 @@ class TreecounterGraphicsText extends Component {
...
                     {/* {this.getTwoWordString(NumberToWords.toWords(planted))} */}
-                    {delimitNumbers(parseInt(planted))}
+                    {planted ? delimitNumbers(planted) : 0}
                   </div>
                 ) : null}
               </span>
             </div>
             {this.props.trillion ||
-            convertNumber(parseInt(community), 2) == 0 ? null : (
+            convertNumber(parseInt(community), 2) === 0 ? null : (
```
norbertschuler: This is the way the native app code is doing it.

PR #1307, opened 2019-08-26
Some more fixes for tree counter numbers

norbertschuler:
- fixed "NaN" value for tree counter details for native clients
- unified null value checks for tree counter numbers
- adapted arrow check (enable link to detail page) for web client which was different from native apps, so the link was almost never shown

### DIFF
--- a/app/components/TreecounterGraphics/TreecounterGraphicsText.js
+++ b/app/components/TreecounterGraphics/TreecounterGraphicsText.js
@@ -65,7 +65,7 @@ class TreecounterGraphicsText extends Component {
                 {this.props.trillion ? (
                   <div>
                     {/* {this.getTwoWordString(NumberToWords.toWords(target))} */}
-                    {target ? delimitNumbers(target) : null}
+                    {target ? delimitNumbers(target) : 0}
                   </div>
                 ) : null}
               </span>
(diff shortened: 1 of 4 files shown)


app/components/TreecounterGraphics/TreecounterGraphicsText.js
```diff
@@ -87,13 +87,13 @@ class TreecounterGraphicsText extends Component {
...
                     {/* {this.getTwoWordString(NumberToWords.toWords(planted))} */}
-                    {delimitNumbers(parseInt(planted))}
+                    {planted ? delimitNumbers(planted) : 0}
                   </div>
                 ) : null}
               </span>
             </div>
             {this.props.trillion ||
-            convertNumber(parseInt(community), 2) == 0 ? null : (
+            convertNumber(parseInt(community), 2) === 0 ? null : (
```
norbertschuler: This is the way the native app code is doing it.

crucialfelix: When we set up the linter it will enforce `===`

meme

crucialfelix:
> adapted arrow check (enable link to detail page) for web client which was different from native apps, so the link was almost never shown

I don't see this in the PR

Why were all the delimitNumbers removed?

norbertschuler:
> > adapted arrow check (enable link to detail page) for web client which was different from native apps, so the link was almost never shown
> 
> I don't see this in the PR

But you added the nice comic to it. https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/d634333efd53c11ecd735dbd56a180d4b66be113/app/components/TreecounterGraphics/TreecounterGraphicsText.native.js#L83 vs. https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/d634333efd53c11ecd735dbd56a180d4b66be113/app/components/TreecounterGraphics/TreecounterGraphicsText.js#L96

> Why were all the delimitNumbers removed?

Because the values were already formatted with `convertNumber()` into something as e.g. "101 Thousend" with could not any more be processed correctly with `delimitNumbers()`.

PR #1318, opened 2019-08-27
Made changes to sort featured project on web in Oldest to Newest orde…

hkurra:
Made changes to sort featured project on web in Oldest to Newest ordering


norbertschuler:
Tomorrow morning I plan to release the version 1.0.6 of the web client. Is this hotfix intended to be a real hotfix, i.d. go to production instantly to fix an unbearable error risking to break something? Then I would try to merge it after the release which would result into a second release of the production version of the web client. The fix for the native apps also included in this pull request is of course too late and will only be merged into the master and develop branch for a next 1.0.7 release of the apps.

app/components/SelectPlantProject/Tabs/featured.native.js
```diff
@@ -25,7 +25,7 @@ export default class FeaturedProjects extends Component {
       }
       return projects;
     }, []);
-    featuredProjects = _.orderBy(featuredProjects, 'id');
+    featuredProjects = _.orderBy(featuredProjects, 'created');
```
crucialfelix: There is already an open pull request that refactors this component. 

https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1287/files#diff-d7f42ea06f612df63b623ba5d07fc473

Should this PR get merged first, I can rebase mine and fix it up.  Or I can just add the change in sort order to that PR.

crucialfelix:
I would say this isn't a hotfix (EMERGENCY).

Also, isn't 'created' likely to have the exact same order as 'id'? Is created a date entered manually on a form, or is it an auto database field?

norbertschuler:
> I would say this isn't a hotfix (EMERGENCY).

@sagararyal what do you think about this. Was this a request from you to do this as a hotfix?

hkurra:
@norbertschuler yes it was a request from Sagar to provide this change as hotfix on the master

sagararyal:
I would merge it under (extraordinary circumstances)

In future we will have to develop an algorithm to determine featured projects.

norbertschuler:
Btw. their is a conflict with the master branch :-(

app/components/SelectPlantProject/Tabs/featured.native.js
norbertschuler: Yes, it would be nice, if you can rebase your PR after this hot fix got merged.

sagararyal:
looks fine to merge

app/components/SelectPlantProject/index.js
```diff
@@ -60,9 +61,8 @@ export default class SelectPlantProject extends Component {
       currencies: { currencies }
     } = props;
 
-    const featuredProjects = plantProjects.filter(
-      project => project.isFeatured
-    );
+    let featuredProjects = plantProjects.filter(project => project.isFeatured);
+    featuredProjects = _.orderBy(featuredProjects, 'created');
```
crucialfelix: This is a mistake. It sorts by 'created' and then directly after this the projects are sorted by price.

norbertschuler: Thanks. That’s why I needed another check. :-)

norbertschuler: Thanks. That’s why I needed another check. :-)

norbertschuler: > This is a mistake. It sorts by 'created' and then directly after this the projects are sorted by price. So the change has no effect.

I thought that are two different arrays: `featuredProjects` sorted by created date and `priceSortedProjects` sorted by price?

crucialfelix: Ah, you are right

PR #1326, opened 2019-08-28
Added Review Designs

harshvitra:
Draft PR for design of Reviews feature. 

Visit the Donate Tab, click on any project and find review button. 
untitled


app/components/PlantProjects/PlantProjectFull.native.js
```diff
@@ -66,6 +68,17 @@ class PlantProjectFull extends React.Component {
               tpoName={tpo_name}
             />
 
+            <TouchableOpacity
```
hkurra: @harshvitra TouchableOpacity is only available on iOS not on android, for that we have our wrapper component TouchableItem under common folder, you should have use that

harshvitra: Please re-check, it works well on both.

hkurra: My point here is u can not apply ripple effect on touchable opacity, if we don't need ripple effect we are good to go

sagararyal:
@harshvitra 
Can't see any changes.
Screen Shot 2019-09-01 at 10 06 33 PM

mahmed0715:
@jmiridis @sagararyal @norbertschuler How do we designed review schema with relation to project and user?
1 to 1 in context on user -> project -> review? this means one user for a certain project there should be  only one single review. 
Currently its adding many more reviews from a single user to same project. I think review should be based on single to single project. Please let us know..

norbertschuler:
@mahmed0715 sorry, I have not yet participated in anything regarding reviews and do not know the idea behind this feature.

sagararyal:
The review is designed to be used by approved users (often foresters or scientists) to write annual reviews or frequent reviews and attach multiple reports. 

I ability to create multiple reviews is according to the concept.

mahmed0715:
ok,

@sagararyal we should show the reviews list by updated date/created date descending order right? @jmiridis need to suplply(its missing currenly)
@jmiridis can you send us updated at and or created at date for a review?

mahmed0715:
@jmiridis we need to have couple of more things in the review object,
To be able to delete/update a review by the logged in user there should be some indication like user id, or what do you think we can disable the buttons for delete/update
There is no indication for review images in the review api.
Thanks

mahmed0715:
@sagararyal @norbertschuler please test the functionality in this PR, 
I will fix codebase and then you can review.

mahmed0715:
Now after all functionality are covered, I am going touse formik to handle input validation and submit with ease.

mahmed0715:
There is a new PR for this, this can be closed or we can keep for only reference

PR #1350, opened 2019-08-30
Fix #1317 by adding missing KeyboardAwareScrollView

### LINKED ISSUE #1317: Free donation amount not visible while entering values on Android device
**Describe the bug**
On an Android device the amount of trees or money to donate to a project is hidden while the keyboard is opened to enter the amount.

both

**To Reproduce**
1. Donate to a project up to the page to choose the amount of trees
2. Click the radio button to donate 150 trees.
3. Click on the input field for the trees to change the amount.

**Expected behavior**
Scroll up the layout to see the input field while typing on the keyboard.

**What actually happens**
The page is not scrolled up, but the keyboard is hiding the input field.

**Screenshots**
Screenshot_20190827-170953

**Smartphone (please complete the following information):**
- Device: Samsung Galaxy S6 Edge
 - OS: Android 7.0
 - Plant-for-the-Planet App Version 1.0.6

norbertschuler:
Fix #1317

When entering the free amount of trees to donate and the keyboard is shown, the form should scroll up on small displays.

### DIFF
--- a/app/components/Currency/TreeCountCurrencySelector.native.js
+++ b/app/components/Currency/TreeCountCurrencySelector.native.js
@@ -10,6 +10,7 @@ import { Dimensions, View, Text } from 'react-native';
 import styles from '../../styles/selectplantproject/selectplantproject.native';
 import NumberFormat from '../Common/NumberFormat';
 import { delimitNumbers } from '../../utils/utils';
+import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view';
 
 class TreeCountCurrencySelector extends React.PureComponent {
   constructor(props) {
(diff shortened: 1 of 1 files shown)


app/components/Currency/TreeCountCurrencySelector.native.js
```diff
@@ -78,96 +79,101 @@ class TreeCountCurrencySelector extends React.PureComponent {
...
-                <Text style={styles.selectedProjectCol}>
-                  {' '}
-                  {i18n.t('label.trees')}
-                </Text>{' '}
-                {i18n.t('label.gift_to', {
-                  name: this.props.giftTreeCounterName
-                })}
-              </Text>
-              {/*<Text>{this.props.giftTreeCounterName}</Text>*/}
+      <KeyboardAwareScrollView
```
norbertschuler: Actually I added just this component - the diff is just so big because I changed the indention of the rest of the code.

crucialfelix: That's good, but we should also maybe redesign the page at the same time. @sagararyal showed me a new design the other day. Maybe that should be done at the same time.

norbertschuler: I won't be able to do it and I would like to have this fix quickly in the next release which should be build next week.

PR #1353, opened 2019-08-30
Feature/pledge events 

harshvitra:
Added Later Button which takes user to previous section


norbertschuler:
1. The text for "e-Wald" is cut on the pledge event detail page:
Screenshot_1567417701

2. The counter is still zero - this might be a backend issue, is it?
Screenshot_1567417769

norbertschuler:
Localization (e.g. with Moment library) for the date of a pledge event was forgotten:

Simulator Screen Shot - iPhone X - 2019-09-02 at 17 12 22

app/components/TreecounterGraphics/FeaturedProjectScroll/FeaturedProject.js
```diff
@@ -4,6 +4,7 @@ import styles from './../../../styles/trillion.native';
 import i18n from '../../../locales/i18n';
 import { greencalendar } from './../../../assets/';
 import { smalltree } from './../../../assets/';
+import moment from 'moment';
```
norbertschuler: I don't know why but mostly I found this code in our app:
```
import moment from 'moment';
import 'moment/min/locales';
```
I don't know if it's necessary to do the second import - maybe someone can explain it to me or read the documentation of Moment.js to find out. At least we should test, that German and English version works with or without it.

Also I initialized the moment library in a components constructor as following:

```
import { getLocale } from '../../actions/getLocale';
...
  constructor(props) {
    super(props);
    moment.locale(i18n.language);
  }
```

sagararyal:
@norbertschuler the 0 are from staging api I think. api is broken

app/components/TreecounterGraphics/FeaturedProjectScroll/FeaturedProject.js
norbertschuler: Did it myself. `import 'moment/min/locales';` seems not necessary any more.

Issue #1293, opened 2019-08-23
Add Icons to download app in the web menu


sagararyal:
Add icons on the Web Menu to download the apps.

These icons need to be localized as well.

<img width="309" alt="Screen Shot 2019-08-23 at 5 40 22 PM" src="https://user-images.githubusercontent.com/5401575/63604866-1868ce00-c5cd-11e9-9572-b73a42aaa5e5.png">

https://zpl.io/awO95rj

`Ignore the Language and Currency dropdown for this task.`

mahmed0715:
can be closed as the pr merged. @norbertschuler

mahmed0715:
@sagararyal is it still pending? I think the pr related to this was merged.

norbertschuler:
should be closed

PR #1354, opened 2019-09-01
Bottom get app icon added on side menu in web only

mahmed0715:
Added get app icon on bottom on web left menu.


app/components/Menu/index.js
```diff
@@ -138,6 +138,20 @@ export default class Menu extends Component {
           </div>
         ))}
         {this.props.userProfile ? this.renderShareButtons() : null}
+        <div className="bottom-banner">
+          <a
+            target="_blank"
+            href="https://apps.apple.com/us/app/plant-for-the-planet/id1444740626"
```
norbertschuler: https://apps.apple.com/app/plant-for-the-planet/id1444740626 also works.

norbertschuler:
Google Badges: https://play.google.com/intl/en_us/badges/
Apple Badges: https://developer.apple.com/app-store/marketing/guidelines/#section-badges

mahmed0715:
#1293

mahmed0715:
@sagararyal please check localized badge addded

mahmed0715:
@sagararyal the size of the google play icon doesnt match with the apple icon(actually not size, but the google icon has a white border around, that looks smaller than apple icon). Please check/match if we are okay to use that?

mahmed0715:
Screenshot from 2019-09-02 23-12-32

sagararyal:
in that case you can download the icons at: https://play.google.com/intl/en_us/badges/
and resize?

app/assets/index.js
```diff
@@ -138,7 +138,10 @@ import forward from './images/forward.png';
 import successAnimated from './images/successAnimated.gif';
 import trees from './images/trees.png';
 import greencalendar from './images/greencalendar.png';
-
+import googlePlayBadge_en from './images/en/google-play-badge.jpg';
```
norbertschuler: Is it only we who think it's weird to use JPG for the Google banner and SVG for the Apple banner? Why not both SVG or both PNG?

mahmed0715: The images in svg are not good enough for both, we used svg for IOS from the package, but got google badge from their site and resized/edited to look similar, thus this happened.

mahmed0715:
@norbertschuler @sagararyal please see now. resized image for google to look similar.

app/components/Menu/index.js
```diff
@@ -7,7 +7,10 @@ import { getLocalRoute } from '../../actions/apiRouting';
 import { context } from '../../config';
 import { allowedUrls } from '../../config/socialShare';
 import { FacebookShareButton, TwitterShareButton } from 'react-share';
+import { getLocale } from '../../actions/getLocale';
 
+let userLang = getLocale() || 'en'; // en|de
+userLang = userLang.split('_')[0];
```
norbertschuler: These above two lines are not necessary as this is already done within `getLocale()`, isn't it?

app/components/Menu/index.js
```diff
@@ -138,6 +141,20 @@ export default class Menu extends Component {
           </div>
         ))}
         {this.props.userProfile ? this.renderShareButtons() : null}
+        <div className="bottom-banner">
```
norbertschuler: If I resize the browser window until the menu does not fit in any more, the banner cover a menu item and if you scroll the menu the banner stay fixed on the menu item they cover before, so you can not select it.

Bildschirmfoto 2019-09-03 um 14 41 16
Bildschirmfoto 2019-09-03 um 14 41 26

norbertschuler: On very small devices, e.g. iPhone 5, the menu is shown as responsive layout, but the banner cover the menu again, so some items can not be selected any more.

Bildschirmfoto 2019-09-03 um 14 45 03

mahmed0715: we should show just under the end of the menu items?not just fixed position(as we do fixed position like at very bottom) @norbertschuler

mahmed0715: @norbertschuler responsive issue solved.

norbertschuler: I think fixing the responsive layout this way is good. The other case where you resize the normal browser window is just an edge case probably no one will experience. So I mark this as resolved.

app/components/Menu/index.js
norbertschuler: I suggest replacing this code with `let userLang = getLocale();`

mahmed0715: changed as requested

mahmed0715:
@norbertschuler please see

PR #1362, opened 2019-09-03
fix: PledgeEvents import connect for updated react-redux 7

crucialfelix:
PledgeEvents https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1324 was merged after the redux update https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1287

There was one import in #1324 that imported using the older API. This updates to the new import path

### DIFF
--- a/app/components/PledgeEvents/PledgeEvents.native.js
+++ b/app/components/PledgeEvents/PledgeEvents.native.js
@@ -8,7 +8,7 @@ import {
   StyleSheet,
   Image
 } from 'react-native';
-import connect from 'react-redux/es/connect/connect';
+import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


app/components/PledgeEvents/PledgeEvents.native.js
```diff
@@ -8,7 +8,7 @@ import {
   StyleSheet,
   Image
 } from 'react-native';
-import connect from 'react-redux/es/connect/connect';
```
norbertschuler: @harshvitra did that change also in https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1353/files

crucialfelix: Ah, now I see it.

PR #1373, opened 2019-09-04
Splash screen for Updates

harshvitra:
Screenshot_1567641215

For now the button is added at Bottom of Trillion Page - 

Screenshot_1567641198

Tried to make it the first page of the app, but app still loaded home page after 1s so redid the change - 

Screen Shot 2019-09-05 at 5 22 40 AM


harshvitra:
To Do - 

- [ ] Open directly before Home Page
- [ ] Detect if Older version and show only if version is old 
- [ ] Add links to open Play Store or App Store

norbertschuler:
Thank, nice, but shouldn't this also be a draft until it's really working, e.g. using the API call to get the outdated information from the backend (which does not yet exist)? (I am sorry, GitHub can not yet convert pull request into draft pull requests.)

PR #1376, opened 2019-09-05
Splash screen for Updates

harshvitra:
Splash Screen for Updates - 

Screenshot_1567640067

To do - 

- [ ] Check if the version is outdated and open this Screen

- [x] Linking App Store and Play Store

Screenshot_1567641198


sagararyal:
@harshvitra 

Spoke with Jorgo, 

It seems you can show that screen when server responds with `426 status code`

it's like 404 but details below:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/426

harshvitra:
This is on hold for some time till we have a response 426 from the backend.

norbertschuler:
> This is on hold for some time till we have a response 426 from the backend.

How should the backend server decide about sending an 426 if it is not knowing the client's version number? We have not API call yet we are giving the client's version number as a payload. See my proposal in #2370 on how we could manage that.

norbertschuler:
replaced with #2466

Issue #1313, opened 2019-08-27
Staging: Button missing to change personal donation target


norbertschuler:
**Describe the bug**
If you want to change your personal treecounter to donate your trees to in the web client you open up a menu by clicking on your profile foto in the upper right corner and then click the button to choose a profile to donate your trees to.
In the dialog opening you have an entry field to find a treecounter, but if I have choosen one no button is visible to confirm my choice.

**Is it on a Production or Development Build?**
staging (develop) 

**To Reproduce**
1. Open menu clicking on your profile image in the upper right corner.
2. Click on choosing profile
3. Find some treecounter to donate to in the dialog
4. Do not see button to confirm your choice.

**Expected behavior**
See a button if you choose a treecounter.

**What actually happens**
No button is visible. The process is stuck.

**Screenshots**
1. personal menu:
<img width="306" alt="Bildschirmfoto 2019-08-27 um 08 48 58" src="https://user-images.githubusercontent.com/1532418/63747852-5e14e780-c8a8-11e9-8d27-9cbdebe980fe.png">
2. dialog to choose a treecounter:
<img width="487" alt="Bildschirmfoto 2019-08-27 um 08 49 08" src="https://user-images.githubusercontent.com/1532418/63747876-6d943080-c8a8-11e9-91f0-c6ea813e282c.png">
3. dialog with choosen treecounter, but not showing button
<img width="483" alt="Bildschirmfoto 2019-08-27 um 08 49 23" src="https://user-images.githubusercontent.com/1532418/63747905-80a70080-c8a8-11e9-986a-f782fc781587.png">

**Desktop (please complete the following information):**
- OS: MacOS latest
 - Browser Chrome latest
 - Plant-for-the-Planet App Version 1.0.5

norbertschuler:
I just recognized it's probably the same problem with the staging API (and database) as in #1217. If I run the web client against the production backend, it all works fine.

norbertschuler:
@jmirides problem is the missing treecounterid as described here https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1217#issuecomment-520233613

norbertschuler:
With the new backend this seems to be solved.

Issue #1217, opened 2019-08-11
Staging: Gift trees to existing user does not work


norbertschuler:
**Describe the bug**
If you choose to gift a tree, choose an existing user from the selector and try to go to the next page to choose a project, the "next" (German "weiter") button does not work.

**Is it on a Production or Development Build?**
development (staging)

**To Reproduce**
1. Go to https://staging.trilliontreecampaign.org/gift-trees
2. Click on field to enter user name
3. Choose an existing user
4. Try to continue to the next page by pressing "next"-button

**Expected behavior**
The list of projects should be shown.

**What actually happens**
Nothing (not even an error)

**Screenshots**
<img width="1091" alt="Bildschirmfoto 2019-08-11 um 14 48 54" src="https://user-images.githubusercontent.com/1532418/62834073-94672b80-bc47-11e9-9602-9b82eb4f4786.png">

**Desktop (please complete the following information):**
- OS: MacOS
 - Browser latest Chrome
 - Plant-for-the-Planet App Version 1.0.3

norbertschuler:
This is not the same bug as #1216 which happened when you does not choose an existing user before.

norbertschuler:
My JavaScript and React knowledge isn't good enough to understand why the `returnValue` is still `undefined` instead of `true` when returning from this `checkValidation` function:

<img width="1680" alt="Bildschirmfoto 2019-08-11 um 16 06 30" src="https://user-images.githubusercontent.com/1532418/62834844-0e9cad80-bc52-11e9-8cac-e095c936606a.png">

Can anyone help?

norbertschuler:
@Jorgo (Nitesh found out for me) in staging we are not getting treecounterId any longer.

Response for https://www.trilliontreecampaign.org/suggest
```
[{"treecounterId":18786,"id":18780,"slug":"barath-rajesh","name":"Barath Rajesh","image":"5b41ac891f21d246723037.png","type":"individual","mayPublish":true,"category":"profile"},{"treecounterId":3094,"id":3094,"slug":"rajesh-a-t","name":"Rajesh A T","image":"5b41ac891f21d246723037.png","type":"individual","mayPublish":true,"category":"profile"},{"treecounterId":127971,"id":113939,"slug":"rajesh-rawal-1","name":"Rajesh Rawal","type":"individual","mayPublish":true,"category":"profile"}]
```
Response for https://staging.trilliontreecampaign.org/suggest
```
[...]

PR #1380, opened 2019-09-06
Fixed: Pick profile button was not visible under dedicate profile mod…

hkurra:
Fixed: Pick profile button was not visible under dedicate profile modal dialog, due to which user were not able to dedicate its trees to anyone

### DIFF
--- a/app/components/Challenge/createChallenge.js
+++ b/app/components/Challenge/createChallenge.js
@@ -79,7 +79,7 @@ export default class Challenge extends Component {
 
   suggestionClicked = (context, event) => {
     this.setState({
-      challenged: event.suggestion.treecounterId,
+      challenged: event.suggestion.id,
       challengedName: event.suggestion.name
     });
   };
--- a/app/components/EditUserProfile/dedicate-trees/ProfilePickerModal.js
+++ b/app/components/EditUserProfile/dedicate-trees/ProfilePickerModal.js
@@ -31,7 +31,7 @@ class ProfilePickerModal extends Component {
     );
   }
   suggestionClicked(event, value) {
-    this.setState({ selectedSuggestion: value.suggestion.treecounterId });
+    this.setState({ selectedSuggestion: value.suggestion.id });
   }
 }
 
--- a/app/components/GiftTrees/index.js
+++ b/app/components/GiftTrees/index.js
@@ -273,7 +273,7 @@ export default class GiftTrees extends Component {
     this.setState({
       form: {
         ...this.state.form,
-        directGift: { treecounter: event.suggestion.treecounterId }
+        directGift: { treecounter: event.suggestion.id }
       },
       giftTreecounterName: event.suggestion.name
     });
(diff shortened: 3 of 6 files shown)


norbertschuler:
Is this the same issue as #1313 which I also referenced in #1217 where I described the problem based on the different backend responses of staging vs. production? I am waiting for a long time now for a backend fix for the staging data, but of course we could also fix the frontend everywhere where this backend calls return data. But then we shouldn't just do it at one but all places.

norbertschuler:
@jmiridis what is your opinion about this - should be rewrite the client software now to get rid of this problem?

hkurra:
@norbertschuler Yes, this is the same bug, the reason of bug is change/deletion  in the property name  of tree-counter object

norbertschuler:
But on the production system this bug does not happen as the property treecounterID isn't missing on production. So it's more a bug of the staging system only.

app/components/EditUserProfile/dedicate-trees/ProfilePickerModal.js
```diff
@@ -31,7 +31,7 @@ class ProfilePickerModal extends Component {
     );
   }
   suggestionClicked(event, value) {
-    this.setState({ selectedSuggestion: value.suggestion.treecounterId });
```
norbertschuler: What about https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/ef2e2ebc077e37600a80d87ea638db8d1d3439e1/app/components/EditUserProfile/dedicate-trees/ProfilePickerModal.native.js#L36?

app/components/GiftTrees/index.js
```diff
@@ -273,7 +273,7 @@ export default class GiftTrees extends Component {
     this.setState({
       form: {
         ...this.state.form,
-        directGift: { treecounter: event.suggestion.treecounterId }
```
norbertschuler: What about https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/ef2e2ebc077e37600a80d87ea638db8d1d3439e1/app/components/GiftTrees/Tabs/GiftUser.native.js#L28?

app/components/Challenge/createChallenge.js
```diff
@@ -79,7 +79,7 @@ export default class Challenge extends Component {
 
   suggestionClicked = (context, event) => {
     this.setState({
-      challenged: event.suggestion.treecounterId,
```
norbertschuler: What about https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/ef2e2ebc077e37600a80d87ea638db8d1d3439e1/app/components/Challenge/Tabs/ChallengeUser.native.js#L83?

norbertschuler:
I am afraid I will not approve this patch (despite my own changes) as on production the resulting values for the property `id` and `treecounterId` are different and can not just be used for each other as a replacement - on production this patch would make everything fail due to wrong treecounter ids - here you see a example response of a `suggest` backend call:

Bildschirmfoto 2019-09-09 um 13 49 03

norbertschuler:
The only solution I see is to fix the database of the staging server and let's dismiss this pull request.

hkurra:
@norbertschuler I am going to close this PR

norbertschuler:
> close this PR

Do it! 👍

PR #1400, opened 2019-09-10
Feature/ChangedHeaderDesign

harshvitra:
Would need more people to test this one, it changes the global header. 

Screenshot_1568067535
Screenshot_1568101604


harshvitra:
> We generally try to put css code in our stylesheet directory, if possible please clean code by moving css to proper files..

Surely will do that wanted to run tests quickly

sagararyal:
@harshvitra 

on iPhone with notches, the menu seems to be taking too much space.

When user scrolls on the page, hide the menu, and only keep the tabs sticking on the top of the screen.

Search UI has to be improved.

Ability to have top bar without header title, that does not stay when the page is scrolled.

harshvitra:
> When user scrolls on the page, hide the menu, and only keep the tabs sticking on the top of the screen.

Closing this, this will be done in incremental steps in all screens

PR #1403, opened 2019-09-10
do not use number name "thousand" any more

norbertschuler:
This would show "100,000", "10,000" or "1,000" instead of "100 Thousand", "10 Thousand" or "1 Thousand".

### DIFF
--- a/app/utils/utils.js
+++ b/app/utils/utils.js
@@ -65,8 +65,14 @@ export function convertNumber(n, d) {
   if (isNaN(n) || undefined) {
     return 0;
   }
-  let x = ('' + n).length - 1;
-  x -= x % 3;
+  let x = ('' + n).length;
+  // use number name starting at millions with 7 digits
+  if (x > 6) {
+    x -= 1;
+    x -= x % 3;
+  } else {
+    x = 0;
+  }
   let p = Math.pow;
   d = p(10, d);
   let rounded = Math.round(n * d / p(10, x)) / d;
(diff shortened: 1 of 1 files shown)


app/utils/utils.js
```diff
@@ -65,8 +65,14 @@ export function convertNumber(n, d) {
   if (isNaN(n) || undefined) {
     return 0;
   }
-  let x = ('' + n).length - 1;
-  x -= x % 3;
+  let x = ('' + n).length;
```
crucialfelix: nitpick:

This would be clearer with names
```js
const numDigits = ('' + n).length;
const digitsInGroup = numDigits > 6 ? (x - 1) % 3 : 0;
```
I have a hard time following the code with just x. It then alters x and re-uses it for something different.

norbertschuler: I had that hard time too understanding the former code first, so I understand your request ;-)

norbertschuler: Actually the formula would read like this to be the same as the original calculation which is also difficult to understand:

`const digitsInGroup = numDigits > 6 ? numDigits - 1 - ((numDigits-1) % 3) : 0;`

app/utils/utils.js
```diff
@@ -61,44 +61,53 @@ export function isAndroid() {
...
-  d = p(10, d);
-  let rounded = Math.round(n * d / p(10, x)) / d;
-  let singular = rounded == 1 ? 1 : 0;
+
+  const numDigits = ('' + number).length;
+  // use number name starting at millions with 7 digits
+  let digitsInGroup = 0;
+  if (numDigits > 6) {
+    digitsInGroup = numDigits - 1;
+    digitsInGroup -= digitsInGroup % 3;
```
crucialfelix: Yes that did confuse me on a quick read.

Viel besser!

PR #1409, opened 2019-09-12
Feature: Refactor progress modal

crucialfelix:
This uses the new redux hooks to connect ProgressModal directly to redux state changes. 

Now the entire top level app doesn't need to re-render every time 
```js
dispatch(setProgressModelState(true))
```
is called.

The modal shows/hides by itself.

### DIFF
--- a/app/components/Common/ModalDialog/ProgressModal.native.js
+++ b/app/components/Common/ModalDialog/ProgressModal.native.js
@@ -1,34 +1,39 @@
-import React, { Component } from 'react';
-import { View, Modal, Alert } from 'react-native';
+import React from 'react';
+import { Alert, Modal, View } from 'react-native';
+import { useSelector } from 'react-redux';
 
+import { getProgressModelState } from '../../../reducers/modelDialogReducer';
 import LoadingIndicator from '../LoadingIndicator';
 
-export default class ProgressModal extends Component {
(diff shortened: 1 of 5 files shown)


app/components/Common/ModalDialog/ProgressModal.native.js
```diff
@@ -1,33 +1,38 @@
...
+  const isOpen = useSelector(getProgressModelState);
+
+  return (
+    <Modal
+      animationType={'slide'}
+      transparent
+      visible={isOpen}
+      elevation="10"
+      onRequestClose={() => {
+        Alert.alert('Loading cancelled in between');
```
crucialfelix: note: this alert was probably for debugging. If the user closed it then they don't need to get an alert telling them that they just closed it.

crucialfelix:
This was a change I had made in react-auth0. I'm trying to strip that PR down so it makes as few changes as possible and only to things directly related to auth0 so that testing and evaluating it is easier.

crucialfelix:
When I submitted large pull requests it seemed that nobody reviewed or took action on them because they it required too much knowledge and work.

So I've tried to submit a really small simple PR here and yet it is still waiting 11 days now.  This is really slowing down team velocity.

Can I get this merged please?

PR #1431, opened 2019-09-16
added proper permission management to access the location

### LINKED ISSUE #1423: Geo-Location does not work on iOS to register trees
**Describe the bug**
When I want to register trees I can not get the location of my iPhone. 

production 

**To Reproduce**
1. Go to 'register trees‘
2. Click on 'get position‘

**Expected behavior**
Get phone location. 

**What actually happens**
Nothing. Not even asking for location permission. 

**Screenshots**
34130D99-D95F-4EFA-8952-40545BE4E447

**Smartphone (please complete the following information):**
- Device: iPhoneX
 - OS: iOS 12.4
 - Plant-for-the-Planet App Version 1.0.7

norbertschuler:
Fix #1423

- Currently iOS does not ask for permission to access the location, but silently fails (can only be activated in the settings to make it work, but nobody tells the user to do that)
- On Android it also already worked before.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -152,6 +152,7 @@ android {
 
 dependencies {
     //implementation project(':react-native-paypal-wrapper')
+    implementation project(':@react-native-community_async-storage')
     implementation project(':react-native-linear-gradient')
     implementation project(':bugsnag-react-native')
     implementation project(':react-native-youtube')
--- a/android/app/src/main/java/org/pftp/MainApplication.java
+++ b/android/app/src/main/java/org/pftp/MainApplication.java
@@ -3,6 +3,7 @@ package org.pftp;
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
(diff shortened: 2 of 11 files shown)


norbertschuler:
In RN 0.60.x the module is extracted into an own module https://github.com/react-native-community/react-native-geolocation and we need to migrate the code to still work.

sagararyal:
@norbertschuler lets not worry about this now, we will have maps soon.

norbertschuler:
@sagararyal so you want me to abandon this pull request?

norbertschuler:
@sagararyal This pull request adds two more native libraries to the Android and iOS project. Therefore you need to update the project with `npm install`. But sometimes I also experience that running `react-native run-ios` afterwards (which is using the Xcode CLI) fails. Then I rebuild the project with Xcode itself and it works well again. (I have to investigate this, maybe the CLI does not really rebuild the iOS app from scratch, but using some cache.)

PR #1433, opened 2019-09-17
Round1: Added eslint error rule 

mahmed0715:
Round: #1335

PR #1442, opened 2019-09-20
Feature/eslint fix  #1353

mahmed0715:
"react/jsx-boolean-value": 1,
"react/jsx-key": 1,
"react/jsx-no-duplicate-props": 1,
"react/jsx-no-target-blank": 1,
"react/jsx-no-undef": 1,
"react-native/no-color-literals": 1,
"react-native/no-raw-text": 1,
"react-native/no-unused-styles": 1

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -64,7 +64,7 @@
     "react/display-name": 0,
     "react/jsx-boolean-value": 2, //
     "react/jsx-key": 2, //
-    "react/jsx-no-duplicate-props": 1, //
+    "react/jsx-no-duplicate-props": 2, //
     "react/jsx-no-target-blank": 1, //
     "react/jsx-no-undef": 1, //
     "react/jsx-sort-props": 0,
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,7 +1,14 @@
 {
-  "extends": ["eslint:recommended", "plugin:react/recommended"],
+  "extends": [
+    "eslint:recommended",
+    "plugin:react/recommended"
+  ],
   "parser": "babel-eslint",
-  "plugins": ["react", "react-native", "json"],
+  "plugins": [
+    "react",
+    "react-native",
+    "json"
+  ],
   "parserOptions": {
     "ecmaVersion": 6,
     "sourceType": "module",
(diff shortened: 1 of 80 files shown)


norbertschuler:
I assume this should replace https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1433 - right?

app/components/Payment/Gateways/StripeCC.native.js
```diff
@@ -88,7 +92,7 @@ export default class StripeCC extends Component {
             this.state.submitClicked ? () => null : () => this.payviaCard()
           }
         >
-          Pay
```
norbertschuler: still a candidate for a new translation resource ;-)

mahmed0715: Done,

app/components/Payment/Gateways/StripeCC.native.js
```diff
@@ -92,7 +91,7 @@ export default class StripeCC extends Component {
             this.state.submitClicked ? () => null : () => this.payviaCard()
           }
         >
-          {PayButtonText}
+          {i18n.t('label.Pay')}
```
norbertschuler: Have you tested that the label name isn't case sensitive as in the resource file I find `"pay": "Pay",`?

mahmed0715: Updated

app/components/Payment/Gateways/StripeCC.native.js
```diff
@@ -91,7 +92,7 @@ export default class StripeCC extends Component {
             this.state.submitClicked ? () => null : () => this.payviaCard()
           }
         >
-          Pay
+          {PayButtonText}
```
crucialfelix: What the linter is asking for is that any text must (for react-native) be wrapped in `<Text></Text>`

By using a variable you may have gotten around the linter error, but the problem is still there.

.eslintrc.json
```diff
@@ -45,20 +52,21 @@
...
       }
     ],
-    "strict": [2, "global"],
-
-    "react-native/no-color-literals": 1,
+    "strict": [
+      2,
+      "global"
+    ],
+    "react-native/no-color-literals": 1, //
```
crucialfelix: You've copied the react-native rules into the overrides, but you should also remove these 4 from the main rules here.  That way they will ONLY run for files named `*.native.js`

app/components/Payment/Gateways/StripeCC.native.js
```diff
@@ -88,7 +91,7 @@ export default class StripeCC extends Component {
             this.state.submitClicked ? () => null : () => this.payviaCard()
           }
         >
-          Pay
+          {i18n.t('label.Pay')}
```
norbertschuler: Sorry, I think you missunderstood me. I just wanted you to use the existing resource `label.pay` which absolutely matching this case: https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/6c5c2fa845b280f6075a188e5b3b405059885af8/app/locales/en/paymentlabels.json#L7

There is no need to add a new resource. Just more work for the translators.

mahmed0715:
> Looks good! I know you've still got a few final lint errors left to solve.

I dont have other errors pending to solve, and without removing the main rules the overrides still works(it overrides the existing rules) and I fixed all of them... @crucialfelix 

Thanks

crucialfelix:
OK now I can see that circleci is incorrectly configured. Those are warnings — it should not fail the check for those. We can fix that.

.eslintrc.json
crucialfelix: Now I see- they are marked as warnings for all code, errors (via the override) for .native files.

This will still show us annoying warnings for no-raw-text in web react files, but let's just leave it for now.

app/components/Payment/Gateways/StripeCC.native.js
crucialfelix: I think this is still a potential runtime error. But this component isn't yet used AFAICT

crucialfelix:
@norbertschuler circle-ci seems to be failing due to just warnings.

```
treecounter-app ❯ npm run lint:errors

> TreecounterApp@1.0.8 lint:errors /Users/crucialfelix/github/treecounter-app
> eslint app --quiet
```

We should merge this as soon as we can.  Thanks!

norbertschuler:
> @norbertschuler circle-ci seems to be failing due to just warnings.

In the CircleCI logs I found this at the end:

```
ERROR in ./web/widgets/NDVI/app/components/App.js

/home/circleci/ttc-app/web/widgets/NDVI/app/components/App.js
  16:52  error  Value must be omitted for boolean attributes  react/jsx-boolean-value
```

Issue #1333, opened 2019-08-29
[JS] Raise eslint error levels and fix errors in code


crucialfelix:
These are really important to lock down the app and improve stability. `eslint` should be able to raise alarms and prevent people shipping many obvious bugs. Pull requests with these bugs will not be able to be merged.

Because changing these touches many files in the codebase, it's important to make changes and file a PR in a single day. Many other developers are working and it is easier and cleaner to do these quickly.

You can read documentation on the individual errors here: https://eslint.org/docs/rules/

Do these one at a time, opening a new PR for each one.

1. In `.eslintrc.json` increase these from warning (1) to error (2)
2. Run `npm run lint:errors` and fix all errors.

- [x]  "no-undef": 1,
- [ ]    "no-unused-vars": 1,

If you are using VSCode then I recommend installing [TypeScript Hero](https://marketplace.visualstudio.com/items?itemName=rbbit.typescript-hero) which includes "Sort and organize your imports (sort and remove unused)". This works on normal .js and .jsx files as well as .ts
It will instantly clean up imports and remove unused.

Unused variables are grayed out in VSCode which makes it very easy to spot and fix.

- [ ]    "no-dupe-keys": 1,
- [ ]    "no-empty": 1,
- [ ]    "no-extra-boolean-cast": 1,
- [ ]    "no-constant-condition": 1,
- [ ]    "no-prototype-builtins": 1,
- [ ]    "no-underscore-dangle": 1,
- [ ]    "no-unreachable": 1,
- [ ]    "no-useless-escape": 1,

crucialfelix:
Hello @amangautam72 any progress on this?

vijayant123:
Hey, I'll be taking a crack at this. If I progress fast enough I'll update you the day when I will create a PR for this. Anyways I will keep you updated on any progress.

Regards,
Vijayant

vijayant123:
#1487 opened a PR. Please review it so I get confidence to proceed further.

norbertschuler:
@vijayant123 I changed your PR description to reference this issue.

PR #1446, opened 2019-09-21
remove unused require statement

norbertschuler:
I just search for 'require' in the code base after I removed some of them with https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1445 and found this last one which seems to be unused and can also be deleted.

Will something like this also be found by eslint?

(code not run yet, needs testing)

### DIFF
--- a/app/components/PledgeEvents/MakePledgeForm.native.js
+++ b/app/components/PledgeEvents/MakePledgeForm.native.js
@@ -46,8 +46,6 @@ import {
 //   ...pledgeSchemaOptions
 // };
 
-let _ = require('lodash');
-
 class MakePledgeForm extends Component {
   state = {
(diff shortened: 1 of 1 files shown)


crucialfelix:
This may have already been fixed in https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1333 but we have not yet seen a PR for it. 

    "no-unused-vars": 2,

In any case, ALL unused variables should be removed (or commented out) and it would be better to do that in a single PR.

mahmed0715:
it has been done with https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1442 first commit.

norbertschuler:
Great, no need to merge this then now.

PR #1461, opened 2019-09-24
Changed names for Pledges to sync with backend

harshvitra:
Me and @jmiridis discussed and changed the naming of Pledges object being used. 

All those who are working on pledge events once this is merged please pull develop. (@sanjayradadiya )

### DIFF
--- a/app/components/PledgeEvents/PledgeTabView.native.js
+++ b/app/components/PledgeEvents/PledgeTabView.native.js
@@ -27,7 +27,7 @@ export default class PledgeTabView extends Component {
     });
   };
   render() {
-    const allpledgeshigh = this.props.pledges.highestPledgeEvents;
+    const allpledgeshigh = this.props.pledges.highestEventPledges;
 
     const highestPledges = allpledgeshigh.map(highestPledges => (
       <View style={styles.tabViewTitleContainer}>
@@ -44,7 +44,7 @@ export default class PledgeTabView extends Component {
       </View>
     ));
 
-    const allpledgesrecent = this.props.pledges.latestPledgeEvents;
+    const allpledgesrecent = this.props.pledges.latestEventPledges;
 
     const latestPledges = allpledgesrecent.map(latestPledges => (
       <View style={styles.tabViewTitleContainer}>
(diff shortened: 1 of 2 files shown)


norbertschuler:
Does this affect any existing API call which is already used in a released app?

norbertschuler:
I just loaded a pledge event on https://staging.trilliontreecampaign.org/ which triggers an API call https://staging.trilliontreecampaign.org/public/v1.3/de/pledgeEvent/esri-user-conference where the result is containing a key now named `latestEventPledges` while the web client is still having code which the old name `{this.props.pledges.latestPledgeEvents.map(pledge => (` and of course the app crashed:

```
TypeError: Cannot read property 'map' of undefined
    at Pledge.value (index.js:97)
```

The native apps does not seem to crash (luckily) but fail to show the latest and highest pledges in the old version any more.

Speaking with Gretas words (not wanting to offending anyone, but it just fits so nicely here): How dare you changing the API signature without increasing it's version, e.g. to 1.4, to be backward compatible with older already released versions of the apps?

app/components/PledgeEvents/PledgeTabView.native.js
```diff
@@ -44,7 +44,7 @@ export default class PledgeTabView extends Component {
       </View>
     ));
 
-    const allpledgesrecent = this.props.pledges.latestPledgeEvents;
+    const allpledgesrecent = this.props.pledges.latestEventPledges;
```
norbertschuler: If you want to changed the names and therefore the signature of an API call, please increase the version of the related API call(s).

harshvitra: @jmiridis lets discuss this, we wanted to make the changes in this version only and not create any higher version (1.4) 

And both the web and app files have been changed and using the same variable names now.

norbertschuler:
@harshvitra @jmiridis we have to do something about this as now the backend already changed the API calls without changing the version number and all backend having that patch crash when opening pledge events, e.g. https://staging.trilliontreecampaign.org/pledge/mandoki-soulmates - so Jorgo should revert his patches and rewrite then the way there exists different versions.

jmiridis:
I will revert the change.

jmiridis:
Change has been reverted.

Issue #1120, opened 2019-07-16
GiftDonations API call is missing data


norbertschuler:
Gift donations are failing in direct gift mode (invitations are fine).
Here an excerpt of a valid request:
```json
{
  "giftMethod": "direct",
  "directGift": {
    "treecounter": 8,
    "message": "sadasda"
  },
  ...
}
```
The request below lacks the `directGift` key which must contain the gifted treecounter.
```json
URI /api/v1.1/de/giftDonationContributions/1
{
  "paymentResponse": {
    "gateway": "stripe_cc",
    "accountName": "stripe_DE",
    "token": {
      "id": "tok_1EvrwRGhHD5xN1UqR5ZbWiBQ",
      "object": "token",
      "card": {
        "id": "card_1EvrwRGhHD5xN1UqD9d1o2lL",
        "object": "card",
        "address_city": null,
        "address_country": null,
        "address_line1": null,
        "address_line1_check": null,
        "address_line2": null,
        "address_state": null,
        "address_zip": null,
        "address_zip_check": null,
        "brand": "Visa",
        "country": "DE",
        "cvc_check": "unchecked",
        "dynamic_last4": null,
        "exp_month": 4,
        "exp_year": 2021,
        "funding": "credit",
        "last4": "9460",
        "metadata": [],
        "name": null,
        "tokenization_method": null
      },
      "client_ip": "[redacted]",
      "created": 1563050135,
      "livemode": true,
      "type": "card",
      "used": false
    }
  },
  "recipientType": "individual",
  "giftMethod": "direct",
  "treeCount": 10,
  "receiptIndividual": {
    "firstname": "[redacted]",
    "lastname": "[redacted]",
    "email": "[redacted]",
    "address": "[redacted]",
    "zipCode": "[redacted]",
    "city": "München ",
    "country": "DE"
  },
  "amount": 10,
  "currency": "EUR"
}
```
user-agent: "Mozilla\/5.0 (Linux; Android 8.0.0; SM-T825) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/75.0.3770.101 Safari\/537.36"

Has happened several times on Juky 13tha and once one on July 6th.

norbertschuler:
The API call is still failing (tested in iOS).

POST-Request https://staging.trilliontreecampaign.org/api/v1.3/de/giftDonation/1/create :
```
{
  "recipientType": "individual",
  "treeCount": 50,
[...]

PR #1467, opened 2019-09-25
Next button stuck issue fix where the trecounter id is not there

mahmed0715:
From #1120 with a users message https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1120#issuecomment-534967409

### DIFF
--- a/app/components/GiftTrees/index.js
+++ b/app/components/GiftTrees/index.js
@@ -289,7 +289,9 @@ export default class GiftTrees extends Component {
     this.setState({
       form: {
         ...this.state.form,
-        directGift: { treecounter: event.suggestion.treecounterId }
+        directGift: {
+          treecounter: event.suggestion.treecounterId || event.suggestion.id
+        }
       },
       giftTreecounterName: event.suggestion.name
     });


norbertschuler:
I am sorry, but this just fixed the wrong suggestion results we retrieve on the staging server where the treecounterId is missing due to wrong mock-up data - compare https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1217

Please test your patch using https://test.trilliontreecampaign.org/ as the backend. I still experience a problem then when trying to call the API `https://test.trilliontreecampaign.org/api/v1.3/de/giftDonation/1/create`.

mahmed0715:
You are right, thats why I kept both in there, you see the code
sugession.treecounterId || sugession.id, 
so in where we have treecounterid it will use that if thats not present in somewhere like staging, we can just keep going with the id.
@norbertschuler

mahmed0715:
And this is not the fix for api call, this is the next button fix reported by a client , see in https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1120#issuecomment-534967409

norbertschuler:
Ok, sorry, I was sure this does not happen on production, but if a user has complained about it, maybe I was wrong.

norbertschuler:
We should close this pull request. I am pretty sure `id` and `treecounterid` are different values on production and that we would use a wrong value to reference the `treecounterid` on production with this patch.

PR #1476, opened 2019-09-26
set pay button enable disable based on loading state

sanjayradadiya:
Pay button would be disabled when the user clicks on it and after a successful payment or failure, it would be enabled again.


norbertschuler:
I'd like to test, but staging does not work, I get an exception trying to pay:
https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1479

Plant-for-the-Planet:
Thanks @sanjayradadiya 
The button works as expected. Can be merged.

PR #1482, opened 2019-09-27
Changed Sidebar Logo

harshvitra:
No more Beta version

Screen Shot 2019-09-27 at 5 54 36 AM


app/assets/index.js
```diff
@@ -26,6 +26,7 @@ import pot from './images/pot.png';
 import loadingBar from './images/loading-bar.gif';
 import tree from './images/darkTree.png';
 import SideMenuImage from './images/side_menu.png';
+import SideMenuImageNew from './images/Planet-Logo.png';
```
norbertschuler: While the `side_menu.png` is 16kb and still is used in the login screen of the native app, the image `Planet-Logo.png` is 151kb and wasn't used in the web client before. It now would also need to be loaded by the user to display our website. It has a size of 1024x1024px as it is used by webpack to generate all the favicons and more by the webpack-plugin `FaviconsWebpackPlugin` during deployement.

So if we want to do proper webdesign, better create a copy of this logo and name it e.g. `side_menu_new.png` and shrink it to the actual size needed below (100x100px) to save traffic and loading time for our users. Or as an alternative solution use the SVG image `Planet-Logo.svg` which is already in the same folder as `Planet-Logo.png` and has only a size of 10kb.

norbertschuler: The SVG image has a shadow which maybe isn't wanted.

norbertschuler: Both image do not have the circle around the logo as in Sagars screen and we still have the problem the the app store button can cover parts of the menu on small screens (but that's another issue):

Bildschirmfoto 2019-09-27 um 11 44 55
Bildschirmfoto 2019-09-27 um 12 03 14

sagararyal:
rebasing to Mater, @norbertschuler please merge to master after review.

PR #1457, opened 2019-09-24
#1304 global currency selector

mahmed0715:
Draft: #1304 
Please check the currency selector, how it work.
@sagararyal

PR #1485, opened 2019-09-27
Feature/1304 language fixed

mahmed0715:
There are some issues with the old PR #1457 
@norbertschuler can you please check this PR against the issues you faced last time


sagararyal:
@harshvitra 
https://zpl.io/awO95rj
can you prettify the design in this PR?

This part.

Screen Shot 2019-09-27 at 11 51 02 PM

norbertschuler:
Two things I would like to test again:
1. Add `?_locale=fr` in the browser to the URL to test what happens if we try to use an unsupported language.
2. Check out how this new part of the menu maybe interfere with the smart banner on small screens.

app/components/Menu/index.js
```diff
@@ -1,16 +1,87 @@
...
+  </Option>
+);
+const singleValue = props => (
+  <SingleValue {...props}>
+    {props.data.icon && <img className="menu-icon" src={props.data.icon} />}
+    <span className="dropdown-label">{props.data.label}</span>
+  </SingleValue>
+);
+
+const customStyles = {
```
norbertschuler: For which reason don't we extract the style into external files?

app/actions/globalCurrency.js
```diff
@@ -0,0 +1,31 @@
...
+
+export function getPreferredCurrency() {
+  if (!cache.currency) {
+    guess();
+  }
+  return cache.currency;
+}
+
+function guess() {
+  // will implement default currency or initialization here
```
norbertschuler: We have a default currency value in https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/b7697d53278e1a9115b4c5c4912715b4b7856efd/app/config/index.js.dist#L19 - do we want to make use of it?

app/components/Menu/index.js
```diff
@@ -139,34 +233,71 @@ export default class Menu extends Component {
...
+          <GlobalCurrencySelector
+            userProfile={this.props.userProfile}
+            updateUserProfile={this.props.updateUserProfile}
+          />
+          <div className="global-selector">
+            <div>
+              <a
+                target="_blank"
+                rel="noopener noreferrer"
+                href="https://apps.apple.com/app/plant-for-the-planet/id1444740626"
```
norbertschuler: It would be great if we could use the Apple App-ID from our meta-tags as the Smartbanner component also does it:
```
  <meta name="apple-itunes-app" content="app-id=1444740626">
  <meta name="google-play-app" content="app-id=org.pftp">
```
So we do not have to store the App-ID twice in our app. The source code of the Smartbanner component we have imported from an external module works like this: https://github.com/patw0929/react-smartbanner/blob/a45823ad085b6e1ee9f0f575fadf4931e7a02be7/src/components/SmartBanner.js#L256

app/components/Menu/index.js
```diff
@@ -139,34 +233,71 @@ export default class Menu extends Component {
...
+                <img src={images['appleStoreBadge_' + userLang]} />
+              </a>
+            </div>
+          </div>
+          <div className="global-selector">
+            <div>
+              <a
+                target="_blank"
+                rel="noopener noreferrer"
+                href="https://play.google.com/store/apps/details?id=org.pftp"
```
norbertschuler: Same as above. It's duplicating the information in the meta-tags used by our Smartbanner component.

app/actions/globalCurrency.js
norbertschuler: (I think it wasn't used before.)

mahmed0715: we never use that one.

app/components/Menu/index.js
mahmed0715: This pr is related to currency + language, Can you please create an issue # for this , I will take care of that.

app/components/Menu/index.js
mahmed0715: Yes please create an issue # for this.

app/components/Menu/index.js
mahmed0715: Please check that these are not direct css class that we use in conjunction with html, these are overriding css. Do we really want to put those in external file? we need to import that css file explicitly here and need to use the classes explicitly.

```const customStyles = {
  control: () => ({
    // none of react-select's styles are passed to <Control />
    display: 'flex',
    width: '100%',
    cursor: 'pointer'
  }),
  container: provided => ({
    ...provided,
    width: '135px',
    display: 'flex',
    cursor: 'pointer'
  }),
  indicatorSeparator: () => ({
    display: 'none'
  }),
  option: () => ({
    display: 'flex',
    padding: '10px',
    width: '100px'
  }),
  menu: provided => ({
    ...provided,
    width: '150px'
  }),
  singleValue: (provided, state) => {
    return {
      ...provided,
      border: 0,
      width: '150px',
      display: 'flex',
      marginLeft: 0,
      color: 'rgba(0, 0, 0, 0.54)',
      fontSize: '17px'
    };
  },
  valueContainer: provided => {
    return { ...provided, padding: 0 };
  }
};```

Comment (mahmed0715) [reply app/components/Menu/index.js]: Okay, we can do that using https://react-select.com/styles#using-classnames

File: app/components/Menu/index.js (hunk 3 above)
Comment (norbertschuler) [reply app/components/Menu/index.js]: Did so: https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1498

File: app/components/Menu/index.js (hunk 4 above)
Comment (norbertschuler) [reply app/components/Menu/index.js]: https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1498

File: app/actions/globalCurrency.js (hunk 2 above)
Comment (norbertschuler) [reply app/actions/globalCurrency.js]: I thought it would be nice to have the default currency extracted from the code, but if it's only in one function, that's also ok for me. But are you sure 'EUR' is the default currency for us? So maybe we should remove `currency: 'USD', ` from the index.js.dist then to avoid irritations?

File: app/actions/getLocale.js [hunk 5]
```diff
@@ -18,11 +19,29 @@ export function getLocale() {
 }
 
 function guessLocale() {
-  let userLang = navigator.language || navigator.userLanguage;
-  let locale = userLang.split('-')[0];
-  if (locale === 'en' || locale === 'de') {
-    return locale;
-  } else {
+  const location = window.location.href.split('/');
+  const _locale = location[location.length - 1];
```
norbertschuler: Maybe we should limit the URL parameter to the supporting languages 'en' and 'de' as http://0.0.0.0:8080/donate-trees?_locale=fr looks pretty ugly:
Bildschirmfoto 2019-09-30 um 12 09 47

norbertschuler:
The smartbanner covers the switches, but I won't fix that as you can easily click it away:
Bildschirmfoto 2019-09-30 um 12 11 15

norbertschuler:
The currency can not be switched any more by the menu dropdown on the page where you can select it together with the amount to donate (and also on the following payment pages). Is this intended behavior?

Bildschirmfoto 2019-09-30 um 12 14 34

app/actions/globalCurrency.js
mahmed0715: Yeah agreed, I am going to use that default from config.js file. thats better to have this consistency.

mahmed0715:
"```The currency can not be switched any more by the menu dropdown on the page where you can select it together with the amount to donate (and also on the following payment pages). Is this intended behavior?```"
yes intended.

mahmed0715:
@norbertschuler we used default currency from config.js file.
and there is a fix for ?_locale=fr we only allow en|de , so it will defaulting to en.

app/actions/globalCurrency.js
mahmed0715: changed as so

app/actions/getLocale.js
mahmed0715: fixed this, defaulting to en

norbertschuler:
> @norbertschuler we used default currency from config.js file.
> and there is a fix for ?_locale=fr we only allow en|de , so it will defaulting to en.

Currently http://0.0.0.0:8080/?_locale=de shows me the English version instead of the German.

norbertschuler:
The menu seems to be at the wrong place (I already cleared my cache):
Bildschirmfoto 2019-09-30 um 14 49 25

mahmed0715:
I just saw this issue is after merging develop into this, There is css that breaks this left menu. Need to fix css.

app/actions/getLocale.js
```diff
@@ -18,11 +19,30 @@ export function getLocale() {
...
+
+  // order of language detection
+  // 1. use language from URL if specified as _locale=de // if not in ['en','de'] we set as 'en'
+  // 2. use default language English if URL contains ?noredirect
+  // 3. use user chosen language from local storage if available
+  // 4. use browser language if possible, but currently only de and en
+  // 5. use English as default language
+  if (_locale.includes('_locale')) {
+    const tempLocale = _locale.split('=')[1];
+    return tempLocale.includes(['en', 'de']) ? tempLocale : 'en';
```
norbertschuler: Can `includes` search the existence of one element of a given array parameter?

mahmed0715:
Left side menu css issue fixed: someone put margin auto for the image on left top, that casuses the issue. just fixed.

app/actions/getLocale.js
norbertschuler: I think this doesn't work and so http://0.0.0.0:8080/?_locale=de shown me the English page.

mahmed0715: My mistake, FIxed

mahmed0715:
Beautification also done with other fixes, @norbertschuler @sagararyal
 @harshvitra no need to come up on this as I have done the beautification

app/components/Menu/index.js
norbertschuler: well, if you do not change this anymore, I don't mind

norbertschuler:
The web client looks o.k. to me now, but I am afraid the native apps crashed as I wanted to check it out - I just opened the app and clicked on the bottom navigation item "Donate":
```
(0, _localStorage.getItemSync) is not a function. (In '(0, _localStorage.getItemSync)('preferredCurrency')', '(0, _localStorage.getItemSync)' is undefined)

guess
    globalCurrency.js:14:31
getPreferredCurrency
    globalCurrency.js:8:9
formatNumber
    utils.js:27:23
```

app/stores/localStorage.js
```diff
@@ -43,7 +43,9 @@ export const fetchItem = key => {
 export const getItem = async key => {
   return window.localStorage.getItem(key);
 };
-
+export const getItemSync = key => {
```
norbertschuler: @crucialfelix can you please check this new function next to `getItem` with your react native expertise. @mahmed0715 introduced it with this patch.

mahmed0715:
@norbertschuler @crucialfelix I tried to fix native app issue with new currency changes, Please verify it works.

norbertschuler:
I am sorry, I still get a crash in the native app (when you click the donate button on one of the projects from the project list):
```
Invariant Violation: Objects are not valid as a React child (found: object with keys {_40, _65, _55, _72}). If you meant to render a collection of children, use an array instead.
    in NumberFormat (created by ConnectFunction)
    in ConnectFunction (at TreeCountCurrencySelector.native.js:145)
```

norbertschuler:
I thought I fixed it all, but I still have an issue choosing the currency in the native app. Only the third row with the amount to change freely changes the currency if I change it in the dropdown:

Simulator Screen Shot - iPhone 11 - 2019-10-02 at 09 50 44

mahmed0715:
Ok i know how to fix this. Will take care of it. @norbertschuler

Issue #1483, opened 2019-09-27
Do not hide support button in responsive view


sagararyal:
Currennt: 

In small screens the support button disappears.

<img width="566" alt="Screen Shot 2019-09-27 at 3 30 27 PM" src="https://user-images.githubusercontent.com/5401575/65760661-231f0180-e13d-11e9-8e09-168e214870cf.png">

Expected: It should not disappear, gift icon can be shown instead.

mahmed0715:
This can be closed as the related PR merged. #https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1486

PR #1486, opened 2019-09-27
Support button and gift icon on small screen issue

mahmed0715:
@sagararyal the gift button only act for logged in user


mahmed0715:
closes #1483

mahmed0715:
Yap, done. now gift icon has same action as the support button @sagararyal

PR #1441, opened 2019-09-19
Prevent web client crash in payment process and show an error instead

norbertschuler:
THIS IS NOT A FIX - IT'S JUST A BETTER INDICATOR OF WHAT MIGHT BE THE PROBLEM!

Fix #1413 
Fix #1439

This is just a draft as it does not really fix the error but just prevents the crash and shows an error instead. 

You can reproduce the error by donating and entering a company address instead of your private address. Then there is neither a `props.donationId` nor a `this.props.paymentStatus` defined. But I have not yet figured out why or where it got lost. Compare the descriptions in the issue #1413 for more.

PR #1489, opened 2019-09-29
Fix crash on gift tree donation crash

mahmed0715:
@sagararyal Reported an issue about gifttree donation crash
Screenshot_20190929-103416_Chrome

Fixed by simple params missing

### DIFF
--- a/app/components/GiftTrees/index.js
+++ b/app/components/GiftTrees/index.js
@@ -541,6 +541,11 @@ export default class GiftTrees extends Component {
                   }
                   amount={this.state.selectedAmount}
                   currency={this.state.selectedCurrency}
+                  paymentDetails={{
+                    amount: this.state.selectedAmount,
+                    currency: this.state.selectedCurrency,
+                    treeCount: this.state.selectedTreeCount
+                  }}
                   expandedOption={this.state.expandedOption}
                   handleExpandedClicked={this.handleExpandedClicked}
                   context={{


mahmed0715:
@sagararyal I tested on test.trilliontreecampaign.org it worked good on my system. No such infinite loading issue with same codebase..

app/components/GiftTrees/index.js
```diff
@@ -541,6 +541,11 @@ export default class GiftTrees extends Component {
                   }
                   amount={this.state.selectedAmount}
                   currency={this.state.selectedCurrency}
+                  paymentDetails={{
```
crucialfelix: I'm going to get the React prop-types lint check in place this week. This error would not be happening if strict prop-types were enforced.

norbertschuler:
@mahmed0715 does this fix our P1 bug #1120 ? If yes, please write "Fix #1120" in the description, so the issue gets closed as soon as this pull request gets merged.

mahmed0715:
Nope, I think #1120 is about BE issue, Jorgo need to check that,
This PR is based on a discussion on Slack, need to create a ticket for this..

mahmed0715:
Sorry Yes it will fix #1120 @norbertschuler

mahmed0715:
closes #1120

norbertschuler:
I am afraid, I still got a crash trying to do the payment for a gift on the test backend with this pull request: https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1497 (I used on of the test stripe credit cards from https://stripe.com/docs/testing#international-cards). My test resulted in an infinite loop of the loading indicator at the following line:

```
CheckoutForm.js:101 Uncaught (in promise) TypeError: Cannot read property 'contribution' of undefined
    at CheckoutForm._this.handlePayment (CheckoutForm.js:101)
```

mahmed0715:
@norbertschuler what can we do to to fix this? Is this because of parameters missing from api call to our BE or its related to payment gateway api?

norbertschuler:
@mahmed0715 I already asked @niteshgrg and @hkurra for help in my PR which should better explain the problem: https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1441 - this PR just prevent the crash, but does not fix the problem of the component `PaymentSelector` (and `CheckoutForm` as a subcomponent) which seem to be used with neither property `paymentStatus` nor property `donationId` correctly set. So somewhere outside of these components the data got lost and I don't know where.

PR #1500, opened 2019-09-30
Added TPO Slug to redirect to treecounter of TPO

### LINKED ISSUE #1261: Link to TPOs from Project Card is broken.
Screen Shot 2019-08-19 at 10 28 13 AM

clicking (by Project Name)

redirects to homepage.

instead it should redirect user to `/t/{slug of tpo}`

harshvitra:
Fixes #1261 
Thank you @jmiridis for quick changes

@norbertschuler it's one of the P1.

### DIFF
--- a/app/components/DonateTrees/index.js
+++ b/app/components/DonateTrees/index.js
@@ -469,6 +469,7 @@ export default class DonateTrees extends Component {
                       expanded={false}
                       plantProject={this.props.selectedProject}
                       tpoName={this.props.selectedTpo.name}
+                      tpoSlug={this.props.selectedProject.tpoSlug}
                       selectAnotherProject
                       projectClear={this.props.plantProjectClear}
                     />
--- a/app/components/GiftTrees/index.js
+++ b/app/components/GiftTrees/index.js
@@ -494,6 +494,7 @@ export default class GiftTrees extends Component {
                     tpoName={this.props.selectedTpo.name}
                     selectAnotherProject
                     projectClear={this.props.plantProjectClear}
+                    tpoSlug={this.props.selectedProject.tpoSlug}
                   />
                 )
               ) : null}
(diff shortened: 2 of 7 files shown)


harshvitra:
`PublicTreecounter` where the user type is tpo we get different output for plantprojects -  userProfile.plantProjects (which does not contain tpoSlug). The component which is rendering the link is - `<TPODonationPlantProjectSelector>` -> `<PlantProjectCarousel>`

There were 5 more instances where I found `<PlantProject>` Component being used, added tpoSlug as prop there

jmiridis:
`app/components/PlantProjects/PlantProjectFull.js`
uses `teaserProps.tpoSlug` line 146
from `this.props.tpoSlug` line88
but `tpoSlug` is not defined as a prop in `PlantProjectFull.js` (should be in line 191) 

How come ESLint doesn'r catch this?

harshvitra:
> `app/components/PlantProjects/PlantProjectFull.js`
> uses `teaserProps.tpoSlug` line 146
> from `this.props.tpoSlug` line88
> but `tpoSlug` is not defined as a prop in `PlantProjectFull.js` (should be in line 191)
> 
> How come ESLint doesn't catch this?

I'm not sure of why it worked without the prop, and we cannot mention it as a required prop because there might be some more components where we don't have the slug to pass as parameter.

norbertschuler:
For displaying projects under a public tree counter this patch fails as there is no `tpoSlug` in the response of `https://test.trilliontreecampaign.org/public/v1.1/de/treecounter/tree-adoption-uganda` in the `plantProject` array property as expected. Instead the slug of the tree counter is directly given as the property `slug` which is not passed down to the components used as you can see if the image below:

Bildschirmfoto 2019-10-07 um 12 06 58

Bildschirmfoto 2019-10-07 um 12 11 34

norbertschuler:
Closed in favor of #1261

Issue #1507, opened 2019-10-03
Currency selector change missing on Gift trees


mahmed0715:
Unfortunately we all missed this gifttrees section with newly introduced currency selector.
Need to do same change as donate trees.

PR #1508, opened 2019-10-03
Currency change on gift tree pages

mahmed0715:
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1507

### DIFF
--- a/app/components/Common/NumberFormat.js
+++ b/app/components/Common/NumberFormat.js
@@ -12,12 +12,18 @@ const NumberFormat = ({
   currency,
   userProfile,
   currencies,
-  globalCurrency
+  globalCurrency,
+  force
 }) => {
   if (!userProfile) {
     userProfile = {
+      // we use redux currency or if not found use localstorage currency
       currency: globalCurrency.currency || getPreferredCurrency()
     };
+  }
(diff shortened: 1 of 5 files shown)


sagararyal:
To me the PR does not allow to select projects.

After selecting the user it takes to donations details page.

mahmed0715:
@sagararyal it would be great if you can give more details about what is  the issue, in details. so that I can come up to fix the issue.

mahmed0715:
Let me clarify about this PR,
In develop branch you see the global currency change reflects on donate tree pages and others but not in gift trees (because we never checked gift trees on it we missed that part), In this PR I have implemented global currency changes that reflects to gift tree pages.

Another Update in this PR is: in develop, you can see that when we are in donate tree pages/currency selector page, the global currency change doesn't affect the local currency selector(not overwriting local currency). 
In this PR I have implemented that overwrite functionality, so now any time you change global currency it will affect all over i.e in the local currency selector, but not vice-versa  i.e local currency change will not affect global currency.

@sagararyal @norbertschuler please check this PR and merge if this two functionality works, if there are other issues that is not related to these two changes please create separate ticket..

Issue #1513, opened 2019-10-04
Sorting Projects Tabs do not work on Android


sagararyal:
On Android, On Donate Flow, Selection List, Price and Country tabs do not sort , and does not behave as expected.

mahmed0715:
We need to have a specific label for RN app/web cleint etc to classify easily @sagararyal 
Is this RN issue with android? @sagararyal

norbertschuler:
I added label for `Web`, `Android` and `iOS` in the same color as `client` if you want to add a more specified label to an issue. We already have issue templates which should make people structure their reports a little bit better. Please use them! ;-)

norbertschuler:
> On Android, On Donate Flow, Selection List, Price and Country tabs do not sort , and does not behave as expected.

@mahmed0715 fixed the price sorting. @sagararyal What else is broken, e.g. for the country sorting? What is the expected behavior?

mahmed0715:
Country sorting is working for ASC, I checked the country codes and they are sorted ascending while we go country tab. [This is based on country code, that might be the case @sagararyal, this looks not sorted with full country name, but its sorted based on country code]
Please let us know what more left on this PR.

mahmed0715:
@norbertschuler can you please merge this PR as this fixes the price sorting issue and I checked the Country soring, its fine(sorted based on country code)..

norbertschuler:
PR is already merged. I kept this issue open as I wanted to ask @sagararyal if anything else needs to be fixed. Otherwise please close this issue.

norbertschuler:
Closing this now.

PR #1516, opened 2019-10-05
fix price country sort android

mahmed0715:
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1513


mahmed0715:
Tested on android devices, Price and country based sorting works.
@sagararyal please check.

PR #1524, opened 2019-10-06
Feature/vodafone signup

harshvitra1412:
Vodafone Users can use the url - https://www.trilliontreecampaign.org/signup/vodafone

The signup page has following changes - 
1. Changing I am a "Vodafone employee"
2. Country option only has Germany
3. Email ID field has validation - only '@vodafone.com and unitymedia.de' allowed 

Screen Shot 2019-10-06 at 11 37 52 PM

Drawbacks which I can't figure out how to solve - 
The validation messages aren't displayed.


app/server/formSchemas/signup.js
```diff
@@ -1255,6 +1255,109 @@ export default {
     submit_url: '',
     submit_method: 'POST'
   },
+  vodafone: {
```
norbertschuler: @jmiridis: How does form validation on server side works. Can this file be manually changed of the client side or is it generated code from the backend?

app/server/parsedSchemas/signup.js
```diff
@@ -31,13 +31,19 @@ const {
   schemaOptions: schemaOptionsOrganization
 } = callParser(schemaLiform.organization);
 
+const {
```
norbertschuler: @jmiridis: same question as above.

web/config/webpack.dev.config.js
```diff
@@ -5,27 +5,27 @@ const path = require('path');
...
-    progressbarwidget: [
-      'babel-polyfill',
-      path.join(__dirname, '../widgets/progressbar/widget.js')
-    ],
-    ndviwidget: [
-      'babel-polyfill',
-      path.join(__dirname, '../widgets/NDVI/widget.js')
-    ]
+    bundle: ['babel-polyfill', path.join(__dirname, '../../index.web.js')]
+    // widget: [
```
norbertschuler: This should not be released.

harshvitra1412: Uncommented this, thank you for finding.

app/server/formSchemas/signup.js
harshvitra1412: I really couldn't understand how the schema and validation worked out but by logical reasoning and some experimentation I got it to work to show for Vodafone.

If someone has complete understanding of this or can direct to any documentation then it would be great.

norbertschuler: So am I. That’s why I asked @jmiridis. 😊

PR #1529, opened 2019-10-09
Turn off deeplink to app: does not work if user has not installed the app

crucialfelix:


crucialfelix:
This would merge into develop though. To get that on master somebody like Norbert needs to merge into both

app/components/App/TreeCounter.js
```diff
@@ -324,10 +324,10 @@ class TreeCounter extends Component {
     if (!this._appRoutes) {
       this.initRoutes();
     }
-    if (window.location.pathname.indexOf('signup') > -1 && this.state.isIOS) {
-      this.openApp(window.location.pathname);
```
norbertschuler: Please check if more code should be reversed or deleted as this was introduced with this commit: https://github.com/Plant-for-the-Planet-org/treecounter-app/commit/7fbb3b91a541250b258e5a1fbdbc1844d149f244

crucialfelix: It doesn't matter. This is the code that is saying "if this is an iOS device and 'signup' is somewhere in the URL then open `trilliontreecampaign://` even though we have no idea if the app is installed"

norbertschuler: ok

PR #1530, opened 2019-10-09
Fix eslint error in index.js.dist which is copied to index.js for circle ci

crucialfelix:


app/config/index.js.dist
```diff
@@ -7,7 +7,7 @@ let userLang = navigator.language || navigator.userLanguage;
  * Parameters that are provided by the server to initialize the App.
  */
 export const initialProps = {
-  locale: userLang === 'en-US' || 'en-UK' || 'en-GB' ? 'en' : null,
+  locale: (userLang || 'en').startsWith('en') ? 'en' : null,
```
norbertschuler: I still think we should just delete this as it is unused as I did in my PR https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1277

crucialfelix: Well I have to run very soon. All PRs are blocked until this is fixed.  You can fix right now by merging this or you can do your own fix.

norbertschuler: I take "merge" ;-)

PR #1539, opened 2019-10-10
TreeCount undefined issue on payment

mahmed0715:
https://test.trilliontreecampaign.org/gift-trees
Go to payment page: see treecount undefined
Screenshot

### DIFF
--- a/app/components/GiftTrees/index.js
+++ b/app/components/GiftTrees/index.js
@@ -545,7 +545,8 @@ export default class GiftTrees extends Component {
                   paymentStatus={this.props.paymentStatus}
                   paymentDetails={{
                     amount: this.state.selectedAmount,
-                    currency: this.state.selectedCurrency
+                    currency: this.state.selectedCurrency,
+                    treeCount: this.state.selectedTreeCount
                   }}
                   expandedOption={this.state.expandedOption}
                   handleExpandedClicked={this.handleExpandedClicked}


app/components/GiftTrees/index.js
```diff
@@ -555,7 +555,8 @@ export default class GiftTrees extends Component {
                     donorEmail: email,
                     donorName: name,
                     plantProjectName: plantProject.name,
-                    giftTreeCounterName: this.state.giftTreecounterName
+                    giftTreeCounterName: this.state.giftTreecounterName,
+                    treeCount: this.state.selectedTreeCount
```
norbertschuler: You're putting `treeCount` in the `context` attribute, but then you also changed to get `treeCount` from the `paymentDetails` attribute below where it already existed before? I think you only need one of these changes, do you?

norbertschuler: The usage of `<PaymentSelector>` in `DonateTrees` and `AppPayments` just added `treeCount` to both `paymentDetails` and `context`.

app/components/Payment/PaymentSelector.js
```diff
@@ -204,7 +204,7 @@ class PaymentSelector extends Component {
               force="true"
             />
           </div>
-          <div>{`${i18n.t('label.trees')}: ${context.treeCount}`}</div>
```
norbertschuler: You don't have to change this after the change above.

mahmed0715: true.

app/components/GiftTrees/index.js
mahmed0715: I tried minimal changes, thus where there is treeCount from context object missing I added,
where there is treeCount in paymentDetail and also you can see in paymentStatus so its not obvious that treecount is passed only once. its in many places for different purposes.. but we can simplify  I think, may be who used that way there were reasons..

app/components/Payment/PaymentSelector.js
mahmed0715: changed

mahmed0715:
Please advice again @norbertschuler

app/components/GiftTrees/index.js
```diff
@@ -545,8 +545,7 @@ export default class GiftTrees extends Component {
                   paymentStatus={this.props.paymentStatus}
                   paymentDetails={{
                     amount: this.state.selectedAmount,
-                    currency: this.state.selectedCurrency,
-                    treeCount: this.state.selectedTreeCount
```
norbertschuler: I would keep this line just in case the component `<StripePayment>` (used as StripeContainer here 👎) which gets forwarded the `paymentDetails` attribute also used that value. Just to keep the implementation the same as other usage of `<PaymentSelector>`. So just revert this chunk and keep the one line change below ;-)

PR #1543, opened 2019-10-13
Fix currency language dropdown style

mahmed0715:
@sagararyal please check this. we tried to implement react select for both..

### DIFF
--- a/app/components/Currency/GlobalCurrencySelector.js
+++ b/app/components/Currency/GlobalCurrencySelector.js
@@ -28,7 +28,8 @@ const customStyles = {
   menu: provided => ({
     ...provided,
     width: '100%',
-    cursor: 'pointer'
+    cursor: 'pointer',
+    fontSize: '17px'
   }),
   indicatorSeparator: () => ({
     display: 'none'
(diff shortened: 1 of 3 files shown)


app/components/Currency/GlobalCurrencySelector.js
```diff
@@ -51,20 +100,46 @@ class GlobalCurrencySelector extends Component {
...
+            value: this.state.preferredCurrency,
+            label: this.state.preferredCurrency
+          }}
+          menuPlacement="top"
+          styles={customStyles}
+          onChange={this.handleCurrencyChange}
+          isSearchable={false}
+          options={currenciesArray}
+        />
+        {/* <div>
```
norbertschuler: We don’t need to keep old commented code. We have Git 😁

mahmed0715: mistake

norbertschuler:
Tested on latest Chrome on MacOS - I am not so convinced by the design, e.g. the language menu looks like it's cut at the right side where the background color suddenly ends with the text instead of filling the whole row:

Bildschirmfoto 2019-10-15 um 12 03 59

Also I think the font size is to large and should be the same as the font in the menu items:

Bildschirmfoto 2019-10-15 um 12 04 07

mahmed0715:
Okay. will change.

mahmed0715:
Fixed those issues. font size is same as other menu items(dont be confused about the capital case for currency, it looks bigger but i checked its same font size)

mahmed0715:
@norbertschuler please review and merge this

Issue #1550, opened 2019-10-16
Warning: Failed prop type: Invalid prop `component` of type `object` supplied to `Route`, expected `function`


norbertschuler:
For every start of the web client, when the whole page is reloaded we got this warning in the console (Chrome and Safari on MacOS):
```
checkPropTypes.js:20 Warning: Failed prop type: Invalid prop `component` of type `object` supplied to `Route`, expected `function`.
    in Route (created by TreeCounter)
    in TreeCounter (created by ConnectFunction)
    in ConnectFunction (created by App)
    in GlobalErrorBoundary (created by App)
    in Provider (created by App)
    in App
```

fllprbt:
Could I take this?

norbertschuler:
Of course

PR #1551, opened 2019-10-16
Upgrade react router

fllprbt:
fixes https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1550


norbertschuler:
Thanks a lot. 🙏

crucialfelix:
```
We'll introduce this API in a minor release (5.x). That means it's 100% backward compatible. (Don't forget, we recently released React Router v5, it was a major version change only because of a dependency version range issue, there were no API changes.)
```

That's good.

PR #1454, opened 2019-09-24
Feature/unfulfilledpledges

harshvitra:
Tasks 
- [x] Design New card for unfulfilled Pledges
- [x] Design new buttons for payment and add more trees
- [x] Add Anonymous option in the form and reflect in the pledges
- [x] Get pre-filled data in the form for logged in users
- [x] Storing pledges/pledge Token of user in local storage 
- [x] Storing token in user profile for logged in user (API not ready)

PR #1555, opened 2019-10-17
Feature/unfulfilled pledge events

### LINKED ISSUE #948: Fix stretched image in pledge event
In Pledge events
https://www.trilliontreecampaign.org/pledge/kirchentag
Images should not be stretched, but rather bound to the circle
Screen Shot 2019-06-21 at 11 07 08 AM

### LINKED ISSUE #1369: Missing translation for pledge events
Some text resources have not been extracted into the resource files in https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/app/components/PledgeEvents/MakePledgeForm.native.js

```javascript
      if (firstname === '') {
        alert('Please Enter First Name');
      } else if (lastname === '') {
        alert('Please Enter Last Name');
      } else if (email === '') {
        alert('Please Enter Email ID');
      } else if (treeCount === '') {
        alert('Please Enter Tree Count');
      }
```

Also normally we do not display alerts with error messages, but show the missing fields with red error message directly next to the field. I think we are using some node modules to create and maintain forms.

norbertschuler:
PR replaces #1454 (in order to better clean up removal of a npm modules with native libraries)

Here is a collections of the issues from my below comments which are not resolvable as the other comments associated with the source code:
- [x] If I make more than one pledge to a single pledge event and then use the button to increase my pledge, only the first pledge always get increased.
- [x] If I make a pledge and navigate to another pledge events, the buttons to donate or increase my pledge are still shown and still related to the first visited pledge event. After visiting a third pledge event, I get again the button to make a new pledge.
- [ ] If you are a logged in user and you are looking the the home page with the global tree counter and the pledge events and then logout, the home page is shown without any pledge event. After reloading the page, the pledge events are shown again.
- [x] If you already have made pledges for two pledge events, go to the first event and increase your pledge, then navigate to the second event and also try to increase your pledge and you will increase the pledge of the first event instead again. (You have to reload the page in order to be able to increase your pledge of the second pledge event.)

Fixes #948
Fixes #1369

When we merge this, we also have to delete the abandoned branch `origin/feature/unfulfilledpledges`.


app/components/PledgeEvents/UnfulfilledPledgeEvent.native.js
```diff
@@ -0,0 +1,188 @@
...
+            </TouchableOpacity>
+          </View>
+          <TouchableOpacity
+            onPress={() => {
+              updateRoute('app_donateTrees', this.props.navigation);
+            }}
+          >
+            <View style={styles.continueButtonView}>
+              <Image
+                style={{ height: '30px' }}
```
norbertschuler: This crashes the iOS-App with `JSON value '30px' of type NSString cannot be converted to a YGValue. Did you forget the % or pt suffix?`

app/components/PledgeEvents/UnfulfilledPledgeEvent.native.js
```diff
@@ -0,0 +1,188 @@
...
+                        </View>
+                    </TouchableOpacity>
+                </View> */}
+
+        <View style={styles.bottomButtonView}>
+          <View style={styles.leftSection}>
+            <Text style={styles.pledgeTreesAmount}>
+              {unfulfilledEvent
+                ? i18n.t('label.treesPledgedAllPledges', {
+                    treeCount: unfulfilledEvent.treeCount.toLocaleString()
```
norbertschuler: I produced a crash here by trying to increase a pledge twice (clicking the button to increase again after I just increased the pledge): 
```
TypeError: undefined is not an object (evaluating 'unfulfilledEvent.treeCount')

This error is located at:
    in MakePledgeForm (created by ConnectFunction)
```

app/containers/Pledge/index.js
```diff
@@ -2,29 +2,55 @@ import React, { Component } from 'react';
...
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.pledges !== this.props.pledges) {
+      console.log(this.props.pledges);
+    }
   }
 
   componentWillUnmount() {
     this.props.clearTimeoutAction(this.props.pledges.timeoutID);
```
norbertschuler: Loading the home page again after I logged out in the web client produced this crash in my browser console:
```
TypeError: Cannot read property 'timeoutID' of null
    at PledgeContainer.componentWillUnmount (index.js:42)
```

app/components/Pledge/index.js
```diff
@@ -22,13 +18,14 @@ let TCombForm = t.form.Form;
...
   return (
-    <div className="pledge-form-layout">
+    <div className="pledge-form-layout pledge-form">
       <div className="row">
         <div className="half">{locals.inputs.firstname}</div>
         <div className="half">{locals.inputs.lastname}</div>
       </div>
       <div>{locals.inputs.email}</div>
       <div className="tree-count">{locals.inputs.treeCount}</div>
+      <div>{locals.inputs.isAnonymous}</div>
```
norbertschuler: I tried to do an anonymous pledge in the native app (not possible on web client!), but that information probably wasn't stored as I it not displayed anonymously afterwards (maybe a backend problem?):

Simulator Screen Shot - iPhone 11 - 2019-10-17 at 11 35 33


Bildschirmfoto 2019-10-17 um 11 36 09

norbertschuler: I saw the API POST request was looking like this - there was no information about being anonymous within the data sent with the request!

# Request

### Endpoint

`POST` https://test.trilliontreecampaign.org/api/v1.3/en/pledgeEvent/gsa-convention/eventPledge

### Data Sent

```json
{
  "firstname": "Norbert",
  "lastname": "Schuler",
  "email": "norbert.schuler@epublica.de",
  "treeCount": "1122"
}
```

app/components/TreecounterGraphics/FeaturedProjectScroll/UnfulfilledEvents.native.js
```diff
@@ -0,0 +1,63 @@
...
+              <View style={{ alignSelf: 'flex-end' }}>
+                <View style={styles.ufpCostView}>
+                  <Text style={styles.ufpCostText}>
+                    {this.props.event.plantProjectCurrency}{' '}
+                    {(
+                      this.props.event.treeCount *
+                      this.props.event.plantProjectTreeCost
+                    ).toLocaleString()}
+                  </Text>
+                </View>
```
norbertschuler: This design does not work for numbers > 1,000:

Simulator Screen Shot - iPhone 11 - 2019-10-17 at 11 47 39

app/containers/Pledge/index.js
norbertschuler: I can not reproduce this. Maybe you have to try doing pledges simultaneously with the web client and the native app or two different browsers. Sometimes I also got this caught error in the console (which does not result in an error for the user):
```
localStorage.js:32 Uncaught (in promise) Error: pledgedEvent not in localStorage
    at localStorage.js:32
    at new Promise (<anonymous>)
    at fetchItem (localStorage.js:29)
    at Pledge.componentDidMount (index.js:61)

```

app/components/Pledge/index.js
```diff
@@ -67,77 +185,401 @@ export default class Pledge extends Component {
...
-                  <span>{i18n.t('label.name')}</span>
-                  <span>{i18n.t('label.trees')}</span>
+                <div className="pledges-header">
+                  {/* <img src={pledge_latest} /> */}
+                  <span style={{ fontWeight: 'bold' }}>
+                    {i18n.t('label.most_recent')}
+                  </span>
+                  <img
+                    src={nextArrow}
+                    onClick={this.openAllPledgesModal}
```
norbertschuler: This opens the same modal dialog showing all pledges sorting with the highest pledges first (as does the link below next to the highest pledges). One would expect the list in this dialog to be sorted by showing recent pledges first:

Bildschirmfoto 2019-10-17 um 12 04 38

app/components/Pledge/index.js
```diff
@@ -67,77 +185,401 @@ export default class Pledge extends Component {
...
+                  <Link
+                    className="right-buttons"
+                    to={getLocalRoute('app_donateTrees')}
+                  >
+                    <img src={nextArrowWhite} className="forward-arrow" />
+                    <div className="donate-text">{i18n.t('label.donate')}</div>
+                  </Link>
+                </div>
+              </div>
+            </div>
```
norbertschuler: These buttons mostly cover some other stuff of the web client, e.g. the logos - can't we make them not be "floating" above the other elements, but be part of the page?

Bildschirmfoto 2019-10-17 um 12 12 52

app/components/Pledge/index.js
```diff
@@ -67,77 +185,401 @@ export default class Pledge extends Component {
...
+          ) : (
+            <p className="make-pledge-button" onClick={this.openPledgeModal}>
+              {i18n.t('label.makePledgeButton')}
+            </p>
```
norbertschuler: This button also covers other elements:

Bildschirmfoto 2019-10-17 um 12 17 33

norbertschuler:
There is one open task by Sagar in the old PR #1454 as following:

> Coming back to home from pledge event does not show the pending pledges, the app needds to be restarted

I could not reproduce that yet, but I can always reproduce the following:

- If you are a logged in user and you are looking the the home page with the global tree counter and the pledge events and then logout, the home page is shown without any pledge event. After reloading the page, the pledge events are shown again.

app/components/Pledge/index.js
harshvitra: Screen Shot 2019-10-17 at 7 45 20 PM

Fixed this but it will have the margin on each page

app/containers/Pledge/index.js
harshvitra: Removed it, was required for testing only

app/components/Pledge/index.js
```diff
@@ -42,20 +39,148 @@ export default class Pledge extends Component {
...
+      myPledge: {},
+      updatingTreeCount: '',
+      pledgeButtonDisabled: false
     };
   }
 
+  componentDidMount() {
+    if (this.props.currentUserProfile) {
+      this.setState({
+        loggedIn: true
```
crucialfelix: `loggedIn` is trivial to compute. Anywhere you need to know if user is logged in just use `!!this.props.currentUserProfile` or a method if you must.

Remove `loggedIn` from state.

app/components/Pledge/index.js
```diff
@@ -42,20 +39,148 @@ export default class Pledge extends Component {
...
+  componentDidUpdate() {
+    if (this.state.loadUserPledges) {
+      if (this.props.entities) {
+        if (this.props.entities.eventPledge) {
+          if (
+            this.props.pledges &&
+            this.props.pledges.allEventPledges &&
+            this.props.pledges.allEventPledges.length > 0
+          ) {
+            if (typeof this.props.entities.eventPledge !== 'undefined') {
```
crucialfelix: Move all this to a method that just returns myPledge or undefined.  Call it in render when and where you need it.

Don't setState, just return the value.

When the redux-triggered fetchPublicPledgesAction completes and sets props then the component will re-render and then you will see myPledge.

What you are trying to do here is to cache something, but it's using state to do it and componentDidUpdate and redux. That's why things are confusing.  State isn't a good place to store memory or cache because it triggers a re-render.  You want the value IN the render, but you don't want to render just because the value changed.

If I were solving this problem I would convert this large component into several functional components. That would make it much easier to maintain and understand.

You could then useMemo to memoize the work in finding myPledge.

https://reactjs.org/docs/hooks-reference.html#usememo

If you don't want to rewrite this whole component then you could memoize it with lodash: https://lodash.com/docs/4.17.15#memoize

app/components/Pledge/index.js
```diff
@@ -67,77 +192,405 @@ export default class Pledge extends Component {
       )[0];
     }
     document.title = getDocumentTitle(selectedPledge.name);
+    let i;
```
crucialfelix: This is kind of weird.

app/components/Pledge/index.js
harshvitra: I need myPledge outside the render function in - onUpdatePledgeSubmit, that's why used setState earlier it was inside render. 

Want to experiment with Hooks but the code was already written by someone else and had to modify things for these features.

app/components/Pledge/index.js
harshvitra: Needed for ranking

harshvitra:
> Sorry, I mixed the terms `donation` and `pledge` - I corrected my first comment above!

Check it commit 9f1ef72 solves it

app/components/Pledge/index.js
crucialfelix: OK, but you could still move this to a method `getMyPledge()` that searches and returns myPledge. Use that anywhere you need it. 

It doesn't have to save this work to state (which causes a re-render) and it doesn't even have to be cached.  It's not doing that much work.

app/components/Pledge/index.js
```diff
@@ -42,20 +39,148 @@ export default class Pledge extends Component {
...
     };
   }
 
+  componentDidMount() {
+    if (this.props.currentUserProfile) {
+      this.setState({
+        loggedIn: true
+      });
+    } else {
+      fetchItem('pledgedEvent').then(data => {
```
norbertschuler: I get these errors in the browser console:
```
localStorage.js:32 Uncaught (in promise) Error: pledgedEvent not in localStorage
    at localStorage.js:32
    at new Promise (<anonymous>)
    at fetchItem (localStorage.js:29)
    at Pledge.componentDidMount (index.js:62)
```
You could catch a failing promise e.g. with adding `.catch(error => console.log(error));`

app/components/TreecounterGraphics/Trillion.native.js
```diff
@@ -101,25 +105,24 @@ class Trillion extends PureComponent {
...
 
-  onMoreClick(id, name) {
-    this.props.selectPlantProjectAction(id);
-    const { navigation } = this.props;
-    //console.log('OnMore');
-    updateRoute('app_selectProject', navigation, null, { titleParam: name });
+    if (this.props.userProfile) {
+      console.log('User Logged in');
+    } else {
+      fetchItem('pledgedEvent').then(data => {
```
norbertschuler: You could catch a failing promise e.g. with adding .catch(error => console.log(error));

app/components/Pledge/index.js
norbertschuler: I thought double negation isn't allowed anymore with eslint now? :-P

app/components/Pledge/index.js
```diff
@@ -42,20 +39,148 @@ export default class Pledge extends Component {
...
+  // Functions for Later/Continue after Pledge Modal
+  openSubmitPledgeModal = () => {
+    this.setState({ SubmitPledgeModalIsOpen: true });
+  };
+  closeSubmitPledgeModal = () => {
+    this.setState({ SubmitPledgeModalIsOpen: false });
+    this.setState({
+      pledgeButtonDisabled: false
+    });
+    setTimeout(window.location.reload.bind(window.location), 2500);
```
norbertschuler: A solution without the need to reload the page would of course be nice.

app/components/Pledge/index.js
```diff
@@ -42,20 +39,148 @@ export default class Pledge extends Component {
...
+  };
+
+  // Functions for Later/Continue after Update Modal
+  openSubmitUpdateModal = () => {
+    this.setState({ SubmitUpdateModalIsOpen: true });
+  };
+  closeSubmitUpdateModal = () => {
+    this.setState({ SubmitUpdateModalIsOpen: false });
+    setTimeout(window.location.reload.bind(window.location), 3000);
```
norbertschuler: A solution without the need to reload the page would of course be nice.

app/components/Pledge/index.js
norbertschuler: The buttons are still floating above the other elements (moving when you change the browser size). So they still can cover other parts of the page. This is more a UX issue as a bug, but I would prefer having these buttons being integrated in the rest of the layout (scrolling with the other elements) rather then now.

Bildschirmfoto 2019-10-18 um 10 32 14

norbertschuler:
> Check it commit [9f1ef72](https://github.com/Plant-for-the-Planet-org/treecounter-app/commit/9f1ef72ab7e3132361c8206fd7645ce718fb3ed9) solves it

This bug still exists: If you already have made pledges for two pledge events, go to the first event and increase your pledge, then navigate to the second event and also try to increase your pledge and you will increase the pledge of the first event instead again. (You have to reload the page in order to be able to increase your pledge of the second pledge event.)

app/components/Pledge/index.js
harshvitra: Refactored the code, Thank you @crucialfelix

app/components/Pledge/index.js
harshvitra: This needs to be sent to the action which decides which API to use.

app/components/Pledge/index.js
harshvitra: Aye Aye captain, but it takes some time for the values to update

harshvitra:
>  If I make more than one pledge to a single pledge event and then use the button to increase my pledge, only the first pledge always get increased.

@norbertschuler I went through this one, only way to fix this is not allow users to pledge more than one pledge, is that fine with you @sagararyal 
Also until contextual donations are not complete we might not be able to detect the fulfillment of pledge. 

> If you are a logged in user and you are looking the the home page with the global tree counter and the pledge events and then logout, the home page is shown without any pledge event. After reloading the page, the pledge events are shown again.

On debugging I found that on Logout we are clearing out redux store and hence the pledge events are also getting deleted.

norbertschuler:
> On debugging I found that on Logout we are clearing out redux store and hence the pledge events are also getting deleted.

Well, one should not implement anything relying on local data in the browser (for any longer then a short user session). Only the data stored in the backend through the API call is persistent. Data stored in the client (browser or native app) is always only cached data which can be lost (if the user changes his browser, clears his browsers data for privacy reasons, changes his mobile device and deletes and reinstalls the app on his phone).

norbertschuler:
> @norbertschuler I went through this one, only way to fix this is not allow users to pledge more than one pledge, is that fine with you @sagararyal

Well, I think one pledge per user per pledge event is enough. If the user wants to pledge more, he can increase his pledge. Not need to make a pledge twice :-)

app/components/PledgeEvents/UnfulfilledPledgeEvent.native.js
norbertschuler: This is still happening.

norbertschuler:
> I think this is low priority, and it should not block the PR. This case does not happen often as well.

I already listed that bug in the PR description at the top. But there is also still a crash when you increase the pledge twice in the native app which is not yet resolved: https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1555#pullrequestreview-303118888

harshvitra:
@norbertschuler  @mahmed0715  Sagar is getting an error after build, but I'm not. Can you test this PR once.

norbertschuler:
I don't get an error when installing and running the web client - web client works, just has minor issues (which could be solved later):

1. I  just get this warning (well, just a warning) in the browser console: 
```
backend.js:6 Warning: react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended, but you can opt-out by setting `ariaHideApp={false}`.
```

2. And if you make a pledge and then logout, still the pledge events does not show until you reload to show them again. Then you got this in the browser console (also not affecting the user experience):
```
Error: pledgedEvent not in localStorage
    at localStorage.js:32
    at new Promise (<anonymous>)
    at fetchItem (localStorage.js:29)
    at Pledge.componentDidUpdate (index.js:88)
    at commitLifeCycles (react-dom.development.js:17349)
    at commitAllLifeCycles (react-dom.development.js:18736)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at commitRoot (react-dom.development.js:18948)
```
If I login again and load the same pledge event, the error is gone away and I can increase my pledge again.

3. Also when the backend API calls are slow it takes a while when you navigate between two pledge events before the state of the buttons (only showing one button if you have not yet made a pledge vs. showing two buttons to donate or increase your pledge) updates so the user is temporarily shown the wrong state.

norbertschuler:
I cannot do testing on MacOS any more until next week, but it should be related to Android as no native stuff was changed if this pull request. I am afraid the Android app crashes when I try to click on an pledge event as following:
```
2019-11-08 16:53:00.125 6186-6237/org.pftp E/ReactNativeJS: Invariant Violation: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    
    This error is located at:
        in PledgeEvents (created by ConnectFunction)
        in ConnectFunction (at SceneView.js:9)
        in SceneView (at StackViewLayout.js:478)
        in RCTView (at View.js:45)
        in View (at StackViewLayout.js:477)
        in RCTView (at View.js:45)
        in View (at StackViewLayout.js:476)
        in RCTView (at View.js:45)
        in View (at createAnimatedComponent.js:151)
        in AnimatedComponent (at screens.native.js:59)
        in Screen (at StackViewCard.js:42)
        in Card (at createPointerEventsContainer.js:26)
        in Container (at StackViewLayout.js:507)
        in RCTView (at View.js:45)
        in View (at screens.native.js:83)
        in ScreenContainer (at StackViewLayout.js:401)
        in RCTView (at View.js:45)
        in View (at StackViewLayout.js:400)
        in StackViewLayout (at withOrientation.js:30)
        in withOrientation (at StackView.js:49)
        in RCTView (at View.js:45)
        in View (at Transitioner.js:141)
        in Transitioner (at StackView.js:19)
        in StackView (at createNavigator.js:57)
        in Navigator (at createKeyboardAwareNavigator.js:11)
        in KeyboardAwareNavigator (at createNavigationContainer.js:376)
        in NavigationContainer (at SceneView.js:9)
        in SceneView (at DrawerView.js:87)
        in RCTView (at View.js:45)
        in View (at DrawerLayoutAndroid.android.js:198)
        in AndroidDrawerLayout (at DrawerLayoutAndroid.android.js:217)
        in DrawerLayoutAndroid (at DrawerView.js:84)
        in DrawerView (at createNavigator.js:57)
        in Navigator (at createNavigationContainer.js:376)
        in NavigationContainer (at AppDrawerNavigatorContainer.js:100)
        in RCTView (at View.js:45)
        in View (at AppDrawerNavigatorContainer.js:99)
        in AppDrawerNavigatorContainer (created by ConnectFunction)
        in ConnectFunction (at TreeCounter.native.js:29)
        in RCTView (at View.js:45)
        in View (at createAnimatedComponent.js:151)
        in AnimatedComponent (at react-native-safe-area-view/index.js:134)
        in SafeView (at withOrientation.js:54)
        in withOrientation (at TreeCounter.native.js:25)
        in App (created by ConnectFunction)
        in ConnectFunction (at App/index.native.js:19)
        in GlobalErrorBoundary (at App/index.native.js:18)
        in Provider (at App/index.native.js:17)
        in App (at renderApplication.js:35)
        in RCTView (at View.js:45)
        in View (at AppContainer.js:98)
        in ChildrenWrapper (at wrapRootComponent.js:9)
        in _class (at wrapRootComponent.js:8)
        in Root (at AppContainer.js:112)
        in RCTView (at View.js:45)
        in View (at AppContainer.js:115)
        in AppContainer (at renderApplication.js:34)
```

harshvitra:
> I cannot do testing on MacOS any more until next week, but it should be related to Android as no native stuff was changed if this pull request. I am afraid the Android app crashes when I try to click on an pledge event as following:

I got that error too but https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1555/commits/10309c587fc30d6b9b4e178f6b46f84f8b6ba708 commit fixed it

harshvitra:
> Also when the backend API calls are slow it takes a while when you navigate between two pledge events before the state of the buttons (only showing one button if you have not yet made a pledge vs. showing two buttons to donate or increase your pledge) updates so the user is temporarily shown the wrong state.

The same happens in app, it is due to 30 secs wait. User will create multiple pledges I'm afraid within those 30 secs

norbertschuler:
Android crash is gone, only two functional issues due to the chosen architecture and slow API calls:

1. For users not logged in after doing a pledge there is still a button "Make a pledge" for some time:

Screenshot_1573229692

2. If you increase the pledge the number is not updated in the increase button, so I could "increase" the pledge to 1118 again until the page got reloaded:

Screenshot_1573229538

harshvitra:
> Android crash is gone, only two functional issues due to the chosen architecture and slow API calls:
> 
> 1. For users not logged in after doing a pledge there is still a button "Make a pledge" for some time:
> 
> Screenshot_1573229692
> 
> 1. If you increase the pledge the number is not updated in the increase button, so I could "increase" the pledge to 1118 again until the page got reloaded:
> 
> Screenshot_1573229538

Fixing this means we need to make the api call every 10 secs or maybe lesser for faster User experience. 

The best solution is server sent events.

norbertschuler:
I will merge this now and have put all the "open" issue into #1610

PR #1568, opened 2019-10-25
Set fall back currency EUR if language is DE

### LINKED ISSUE #1563: Fallback currency depending on user Location
From @sagararyal 
We need to fallback the default currency depending on the location of the user.
Example: if user in Germany, either language is German/English we need to set currency EUR.
If India we will set language English and currency as INR

@norbertschuler what do you think.

mahmed0715:
We will do vast implementation based on user location later. for now we set EUR if language is DE. 
@sagararyal please check this.

Fix #1563

### DIFF
--- a/app/actions/globalCurrency.js
+++ b/app/actions/globalCurrency.js
@@ -1,6 +1,7 @@
 import { saveItem, getItemSync } from '../stores/localStorage';
 import { setCurrency } from '../reducers/currencyReducer';
 import { context } from '../config';
+import { getLocale } from './getLocale';
 let cache = { currency: '' };
 
 export function getPreferredCurrency() {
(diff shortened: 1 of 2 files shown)


app/actions/globalCurrency.js
```diff
@@ -11,7 +12,11 @@ export function getPreferredCurrency() {
 }
 
 function guess() {
-  cache.currency = getItemSync('preferredCurrency') || context.currency;
```
norbertschuler: What about the native apps? https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/a42b62adfd32e14325203be1da702f07564e20fc/app/actions/globalCurrency.native.js#L13

sagararyal: this hasn't yet been implemented in native apps

norbertschuler: Ok, but I thought if we have the same code also already in the native app, we would have to adapt it there as well. But I haven't yet checked deeper, so @mahmed0715 can decide best if the linked code also should be adapted.

mahmed0715: I think its okay to merge this for now. Ultimate result is same for both. 
BTW I am planning a location based currency set mechanism now. That will be adapted both native and web end.

PR #1572, opened 2019-10-27
Added global location based currency and language mechanism

mahmed0715:
@sagararyal Now we have a mechanism to set location based currency and language 
For now language wont work because we need to have a dynamic mechanism to add all the translations for all languages. we can do that not just manually. need to think @norbertschuler 
Its good to merge, phrase 1 done. 

@jmiridis need to update currency api ```https://test.trilliontreecampaign.org/public/v1.1/en/currencies```
We do not need the currency names in the response anymore, and the rates doesnt support all currencies, its supports about 85 currency. My local currency doesnt support. ```https://docs.openexchangerates.org/docs/supported-currencies``` has some more supported currencies..


app/utils/api.js
```diff
@@ -130,7 +130,17 @@ export async function postDirectRequest(path, data, authenticated = false) {
     .then(onAPIResponse)
     .catch(onAPIError);
 }
-
+export async function getExternalRequest() {
```
norbertschuler: Can you name this something less common, e.g. getDataFromIPStack()?

app/actions/fetchLocation.js
```diff
@@ -0,0 +1,26 @@
+import { getItemSync } from '../stores/localStorage';
+import { getExternalRequest } from '../utils/api';
+import countryCodes from '../assets/countryCodes.json';
+import supportedCurrency from '../assets/supportedCurrency.json';
```
norbertschuler: This lists are nice but every now and then they probably need to be updated, so we (1) have to remember it if we have no automated process and (2) please reference your source you got the data from e.g. by adding a comment with the source.

mahmed0715: Those are static data, no need to update, thats why we kept in assets.
The dynamic part that changes everyday is Rates, we fetches that from api.
From out current implementation ```https://test.trilliontreecampaign.org/public/v1.1/en/currencies``` we got the supported currencies, I am working on something so that we call an api that supports about 170 currency(currently we support only 85 currency)  and then I will update the supported currency list. Those are static data.

app/utils/api.js
mahmed0715: This is part of a plan, I intended to use this as a endpoint to call external get request, like a generic method.
Let me update code so that you can see what I planned.

mahmed0715: for now We are using ipstack may be next time we might use some other service, that time we just need change the api_key and the endpoint url...

app/actions/fetchLocation.js
norbertschuler: Yes, countries and currencies are quite static, but also can change in years. That why I requested to add a comment of the source you copied them from (also that could change as well in this time scale).

mahmed0715: Annotation added. The supported currencies list might change only when we change the api.

app/config/index.js.dist
```diff
@@ -22,5 +22,7 @@ export const context = {
   },
   ios:{
     appId: '1444740626'
-  }
+  },
+  locationApikKey: '1dfab408903c332b660f5e7cb90da06a',
```
norbertschuler: Have your registered at https://ipstack.com/product to get this API key? Normally @sagararyal registers with some official Plant-for-the-planet.org email for the services we use and shares the accounts with 1Password. @sagararyal have you done this?

app/config/index.js.dist
```diff
@@ -22,5 +22,7 @@ export const context = {
...
+  currencyApiKey: 'd581a5406d694920aa493b9143a9e257'
```
norbertschuler: Same question as above for https://openexchangerates.org/signup? @sagararyal have you also registered Plant-for-the-Planet to get an API key here and added the account to 1Password?

app/utils/api.js
```diff
@@ -130,7 +130,14 @@ export async function postDirectRequest(path, data, authenticated = false) {
     .then(onAPIResponse)
     .catch(onAPIError);
 }
-
+export async function getExternalRequest(params) {
```
norbertschuler: O.k., I understood your intention about this now ;-) How about putting this function to the end of this file to separate it from the other API calls to our backend?

app/config/index.js.dist
mahmed0715: Yes I registered a test account there using my email address and the Key's are in the config file for testing.

app/utils/api.js
mahmed0715: done

mahmed0715:
Sure, Lets wait for @sagararyal gets us official api keys/app_ids.

mahmed0715:
@sagararyal @norbertschuler The phrase 2 also done in this PR.
There are changes in everywhere related to currency/rates/price etc. 
Please check ...

app/server/routes/fos_js_routes.json
```diff
@@ -1,6741 +1,5390 @@
 {
-    "base_url": "",
```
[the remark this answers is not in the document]
mahmed0715: One thing we can define for our project is: indentation format
i.e space based indentation with 2 space
that way all developers should set the rule in the code editor and that way we dont need to worry about reformatting..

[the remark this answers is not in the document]
mahmed0715: for me this file has a change for the currency api...

app/config/index.js.dist
```diff
@@ -22,5 +22,6 @@ export const context = {
   },
   ios:{
     appId: '1444740626'
-  }
+  },
+  locationApikKey: '1dfab408903c332b660f5e7cb90da06a'
```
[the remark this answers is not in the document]
mahmed0715: Yes, its not the official api key, i created this for test purpose, that I think its okay we run with it in staging/test if you want to. @sagararyal might get us the key on weekend...

app/components/Currency/CurrencySelector.js
```diff
@@ -3,15 +3,14 @@ import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import TextBlock from '../Common/Text/TextBlock';
 import i18n from '../../locales/i18n.js';
-import { currencySort } from './utils';
```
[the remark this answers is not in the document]
mahmed0715: true. will do.

app/config/index.js.dist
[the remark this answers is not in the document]
norbertschuler: Unfortunately we haven't yet any concept to use different API key (enviroments) for develop or production other then manually changing the values in our `config/index.js` file. (So when we want to do a release we really have to thoroughly check the content of this file before building the client for production which is a source of errors.) - but that's not the problem of this PR :-)

app/server/routes/fos_js_routes.json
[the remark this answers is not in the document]
norbertschuler: @mahmed0715 as we update this file from the backend repository when it probably is auto-generated by some PHP code, that code would have to be adapted too. Otherwise the backend format should be ruling this file.

norbertschuler:
@mahmed0715 also if you do not write this patch for the native app, though it's sharing code with the web client and currently is crashing when you try to do a donation (I assume it's crashing in the `convertCurrency()` method):
```
TypeError: undefined is not an object (evaluating 'currencies.rates[fromCurrency]')

This error is located at:
    in TreeCountSelector (at TreeCountCurrencySelector.native.js:163)
    in RCTView (at View.js:45)
    in View (at TreeCountCurrencySelector.native.js:162)
    in RCTView (at View.js:45)
    in View (at TouchableItem.native.js:33)
    in TouchableItem (at Card/index.native.js:7)
    in CardLayout (at TreeCountCurrencySelector.native.js:88)
    in RCTScrollContentView (at ScrollView.js:976)
    in RCTScrollView (at ScrollView.js:1115)
    in ScrollView (at KeyboardAwareHOC.js:509)
    in KeyboardAwareScrollView (at TreeCountCurrencySelector.native.js:84)
    in TreeCountCurrencySelector (at DonateTrees/index.native.js:289)

convertCurrency
    index.bundle?platform=ios&dev=true&minify=false:169292:47
getSingleTreeRate
    index.bundle?platform=ios&dev=true&minify=false:190380:46
calculateAmount
    index.bundle?platform=ios&dev=true&minify=false:190370:61
proxiedMethod
    [native code]:0
TreeCountSelector
    index.bundle?platform=ios&dev=true&minify=false:191483:48
TreeCountSelector
    [native code]:0```

app/config/index.js.dist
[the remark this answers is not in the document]
mahmed0715: Yah the api key is different than the PR solves the issue.
Yah we must check the config file before we build for production, 
I meant that if you are okay with the PR's code and functionality we can move forward and merge to develop if you want to..

mahmed0715:
@sagararyal The plan is to make these thing simpler than before and faster,

We changed the api and the structure is changed(@jmiridis knows why, its because to less dependency + less data transfer + simplicity in calculation)(currently we are much dependent on the current api that is much static data and unnecessary complex structure) 

In backend @jmiridis call the new api once a day using cron and cache it,

To enable/disable currency dependent on paypal+stripe we have a config file in https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/feature/currency-by-location-1563/app/assets/supportedCurrency.json  we can update like this ``` {
        "Symbol": "AED",
        "Name": "United Arab Emirates Dirham",
        "enabled": true
        "stripe":true//if needed
        "paypal": false//if needed
    }```

Another concern was how frequently we change this information? I guess not more than once a year depending on paypal or stripe changes their supported currency(They don't change that frequently), right?

app/components/Currency/CurrencySelector.js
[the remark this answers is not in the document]
norbertschuler: don't forget about it :-)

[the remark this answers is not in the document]
mahmed0715: done

mahmed0715:
@norbertschuler the crash issue solved in native. Please check.

sagararyal:
@mahmed0715 we can't use USD as base currency and convert all currencies based on that. Because at many cases exchange rates are pegged to either USD or EURO.

Using base currency will lead to either inflate or deflate the exchange rate.
Please explain how the change has been made.

mahmed0715:
We convert everything to usd tgen convert to other currency. I have checked couple of currencies converting this way . There is a very tiny diff thats not a big deal, if you know some specific currencies we can not use this way please let me know i will find a solution to this issue.. @sagararyal

norbertschuler:
@mahmed0715 I check that native crash is fixed and the native app does work now. I merged develop so resolve the conflicts with the latest version. 
Now you just have to clarify if handling currency conversion is ok for @sagararyal.  
Also I would like to know if @jmiridis now has removed the usage of fixer.io in the backend in favor or using openexchangerates.org or if we use both?

mahmed0715:
@norbertschuler this can be closed and there is a new PR which uses old currency api. No change in calculation @sagararyal

mahmed0715:
closing as there is a reference to this in the new PR.

Issue #1576, opened 2019-10-30
Donation amount is cut for large number of donated trees


norbertschuler:
**Describe the bug**
The field for the donation amount when freely entering a number of trees to donate is not large enough for big spenders :-) So the amount can be cut.

**Is it on a Production or Development Build?**
production

**To Reproduce**
1. Go to 'Donates trees'
2. Click on 'Donate'
3. Click on 'Next'
4. Enter a large number of trees to donate in the input field in the third row.
5. See error (amount of money is cut)

**Expected behavior**
We should provide enough place in the layout to also display huge amounts of donations.

**What actually happens**
The text for the donation amount is cut.

**Screenshots**
<img width="834" alt="Bildschirmfoto 2019-10-30 um 09 55 31" src="https://user-images.githubusercontent.com/1532418/67843184-c926a880-fafb-11e9-9fed-c2ab0a88294c.png">

**Desktop (please complete the following information):**
- OS: latest MacOS
 - Browser latest Chrome
 - Plant-for-the-Planet App Version 1.1.0

mahmed0715:
Yah I also noticed that, its a simple css change.

PR #1578, opened 2019-10-30
Do not cut donation amount for a large number of trees

quique:
Closes: #1576

### DIFF
--- a/web/stylesheet/donateTrees/_donate_trees.scss
+++ b/web/stylesheet/donateTrees/_donate_trees.scss
@@ -142,7 +142,7 @@
         border: none;
         border-radius: 0;
         padding: 0;
-        width: 60px;
+        width: 80px;
         border-bottom: 1px solid mix(white, #424242, 70%);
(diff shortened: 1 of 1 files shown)


web/stylesheet/donateTrees/_donate_trees.scss
```diff
@@ -142,7 +142,7 @@
         border: none;
         border-radius: 0;
         padding: 0;
-        width: 60px;
+        width: 80px;
```
norbertschuler: Already better then before. The challenge is to find a nice layout working on big screens as well as on mobile devices. I tested setting `width: 45%;` here, removing the width attribute from `.price-conversion__equal` and setting `width: 45%;` for `price-conversion__radio`, but that also has issues on small screens.

norbertschuler: Please let us know if you like to try to make any more changes for this pull request.

norbertschuler:
replaces by #1629

Issue #104, opened 2018-06-11
Pages are not all responsive on small screens.


bansallakshi:
Update: The site is already partially responsive, but some page still have problems, e.g.

1 . The project liste if you want to donate and scroll down below the featured projects:

<img width="225" alt="Bildschirmfoto 2019-11-01 um 09 12 16" src="https://user-images.githubusercontent.com/1532418/68011517-d7093480-fc87-11e9-9d3f-dce1039147c6.png">

2. The "modal" display of a chosen project from the above list before you select it to display its details:

<img width="225" alt="Bildschirmfoto 2019-11-01 um 08 58 05" src="https://user-images.githubusercontent.com/1532418/68011587-f99b4d80-fc87-11e9-9760-49cf596df012.png">

Maybe there are also other pages with problems on small screens.

norbertschuler:
There are still non responsive pages, e.g. this "modal" project dialog:

![Simulator Screen Shot - iPhone X - 2019-09-23 at 16 26 04](https://user-images.githubusercontent.com/1532418/65434578-51d86780-de1f-11e9-983f-800e86185d13.png)

sagararyal:
Closing this, we'll redesign the web

PR #1583, opened 2019-11-01
adding in responsive styling to fix overlapping on smaller resolutions

sgluzins:
Fix for Issue #104 

Note: This was tested to fix the issue on web.

### DIFF
--- a/web/stylesheet/manageProjects/_manage_project.scss
+++ b/web/stylesheet/manageProjects/_manage_project.scss
@@ -1,6 +1,22 @@
+$breakpointTablet: 769px;
+$breakpointDesktop: 1130px;
+
+@include for-tablet {
+  .app-container__content {
+    width: 100%;
+  }
+}
+
+@media only screen and (min-width: $breakpointTablet) and (max-width: $breakpointDesktop) {
+  .app-container__content {
+    width: 53%;
+    margin-left: auto;
+    margin-right: auto;
+  }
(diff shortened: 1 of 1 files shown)


web/stylesheet/manageProjects/_manage_project.scss
```diff
@@ -1,6 +1,22 @@
...
+}
+
+@media only screen and (min-width: $breakpointTablet) and (max-width: $breakpointDesktop) {
+  .app-container__content {
+    width: 53%;
+    margin-left: auto;
+    margin-right: auto;
+  }
+}
+
```
norbertschuler: Thank you very much for your contribution. This file is actually used for defining the styles of the page where tree planting organizations manage their projects (adding new ones and editing existing ones). It's not the page where users select and look at the projects. I applied this pull requests locally, but still have problem to look at our projects on small screen (maybe the images are just to big):

Bildschirmfoto 2019-11-01 um 08 58 05

If you are interested in digging deeper in our styles and participate in the project that would be great. Maybe you know how the already existing styles in https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/web/stylesheet/_mixin.scss affect all other pages to identy the problem on the page in my above screen shot (I am myself not very skilled with CSS or SCSS).

norbertschuler:
closed as it probably does not help in this way

PR #1587, opened 2019-11-02
Map integration

sanjayradadiya:
I have used react native map library and set on both single and multi-tree selection 

With this pull request we moved the Android credentials to `local.properties`, so they are not any more part of the public sources.


package-lock.json
```diff
@@ -1270,6 +1270,45 @@
         "core-js": "^2.5.7"
       }
     },
+    "@mapbox/geo-viewport": {
```
norbertschuler: In the latest version of the develop branch I removed `package-lock.json` from the project as it shouldn't be part of it anymore (it's already ignored by .gitignore, so I should not be added again). Everyone should just generate it locally in his own workspace.

norbertschuler:
@sanjayradadiya I saw you decided to remove `@react-native-mapbox-gl/maps` again. If it isn't needed and you therefore didn't need to update the native projects, I suggest you complete revert all changes still there in the `ios` and `android` folders.

sanjayradadiya:
(React Native all versions) Specify your Google Maps API Key:

Add your API key to your manifest file (android/app/src/main/AndroidManifest.xml):
```
<application>
   <!-- You will only need to add this meta-data tag, but make sure it's a child of application -->
   <meta-data
     android:name="com.google.android.geo.API_KEY"
     android:value="Your Google maps API Key Here"/>
</application>
```

sanjayradadiya:
image

PR #1613, opened 2019-11-08
API call issue fixed

harshvitra:
Issue fixed by keeping the API call at 30 secs and calling the API again when the user Makes or Updates the pledge. 

The solution was rather simple :-P we over engineered. 

Tested in App and works perfectly, faster than before as it doesn't even wait for 10 secs, cannot test in Web but still coded blindly for now. I hope it works.


app/components/Pledge/index.js
```diff
@@ -347,7 +281,7 @@ export default class Pledge extends Component {
             isOpen={this.state.AllPledgesModalIsOpen}
             pledges={pledges}
             closeAllPledgesModal={this.closeAllPledgesModal}
-            name={selectedPledge.name}
+            name={pledges.name}
```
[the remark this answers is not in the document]
norbertschuler: O.k., now I am just getting the logged error in the console again - no more crash:
```
Error: pledgedEvent not in localStorage
    at localStorage.js:32
    at new Promise (<anonymous>)
    at fetchItem (localStorage.js:29)
    at PledgeContainer.componentDidUpdate (index.js:63)
```

norbertschuler:
Another bug (edge case): Try looking at a pledge event and then logging out. Then try to make a pledge as a logged out user and you won't be able to do a pledge and instead get the message `{code: 401, message: "Invalid JWT Token"}` as a response from the API call. You will have to completely reload the page to be able to send a pledge as a logged out user.

harshvitra:
Wait give me a minute

harshvitra:
>  Error in browser console:
`backend.js:6 Warning: react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended, but you can opt-out by setting `ariaHideApp={false}`

fixed this too, please merge this also

norbertschuler:
Another bug (edge case): Try looking at a pledge event and then logging out. Then try to make a pledge as a logged out user and you won't be able to do a pledge and instead get the message `{code: 401, message: "Invalid JWT Token"}` as a response from the API call. You will have to completely reload the page to be able to send a pledge as a logged out user.

app/actions/pledgeAction.js
```diff
@@ -149,9 +150,9 @@ async function getLocalStorageItem(key, res) {
   }
 }
 
-export function clearTimeoutAction(id) {
+export function clearTimeoutAction(timeoutID) {
   return dispatch => {
-    clearInterval(id);
-    dispatch(clearTimeoutID());
+    clearInterval(timeoutID);
+    dispatch(clearTimeoutID(timeoutID));
```
harshvitra: We don't need TimeoutID to clear it -> 

```
[clearTimeoutID]: (state) => {
      return { ...state, timeoutID: null };
    }
```

We are setting it to null so no need of any payload data

norbertschuler: O.k., you're right, I reverted my unnecessary commit.

Issue #1100, opened 2019-07-13
Deleting trees does not update My Trees & Tree Counter


sagararyal:
When a tree is deleted the tree counter should update to reflect the changes..

vid_bug: 11:00 - 11:30

norbertschuler:
Please test again, should be fixed?

PR #1624, opened 2019-11-11
Validation fixed for register tree issue #1577

### LINKED ISSUE #1577: Register trees: Don't show errors for valid input if other input is missing
**Describe the bug**
If you do not specify a location for registering a tree, but already entered a valid date, next to the error for specifying a location also an error for the date is shown.

production and develop

**To Reproduce**
1. Go to 'Register trees'
2. Enter a valid date
3. Click button to register

**Expected behavior**
Just show an error for the missing location, but not for the date

**What actually happens**
Showing errors for the date and the location.

**Screenshots**
Bildschirmfoto 2019-10-30 um 12 04 40

**Desktop (please complete the following information):**
- OS: latest MacOS
- Browser latest Chrome
- Plant-for-the-Planet App Version 1.1.0

rabiul042:
I have fixed the form validation problem and parameter passing issue through the API of single and multiple trees.

Fix #1577

### DIFF
--- a/app/components/RegisterTrees/index.js
+++ b/app/components/RegisterTrees/index.js
@@ -139,7 +139,12 @@ export default class RegisterTrees extends Component {
             : null;
     }
 
-    this.props.onSubmit(this.state.mode, null, plantProject);
+    let formvalue =
+      this.state.mode == 'single-tree'
+        ? this.state.formValueSingle
+        : this.state.formValueMultiple;
+
+    this.props.onSubmit(this.state.mode, formvalue, plantProject);
     event.preventDefault();
   }
 
--- a/app/containers/RegisterTrees/index.js
+++ b/app/containers/RegisterTrees/index.js
@@ -14,7 +14,6 @@ import {
   schemaOptionsSingleTree,
   schemaOptionsMultipleTrees
 } from '../../server/parsedSchemas/registerTrees';
-import { handleServerResponseError } from '../../helpers/utils';
 
 class RegisterTreesContainer extends PureComponent {
   constructor() {
@@ -27,14 +26,7 @@ class RegisterTreesContainer extends PureComponent {
   }
 
   onSubmit = (mode, registerTreeForm, plantProject) => {
-    /*registerTreeForm =
-      registerTreeForm || this.refs.registerTrees.refs.registerTreeForm;
-    // console.log(registerTreeForm.validate());
-    let value = registerTreeForm.getValue();
-    value = Object.assign({}, value);
-    value = mergeContributionImages(value);*/
-
-    let result = this.refs.registerTrees.refs.registerTreeForm.validate();
+    this.refs.registerTrees.refs.registerTreeForm.validate();
 
     let value = Object.assign({}, registerTreeForm);
     value = mergeContributionImages(value);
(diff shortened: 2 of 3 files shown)


app/components/RegisterTrees/index.js
```diff
@@ -139,7 +139,12 @@ export default class RegisterTrees extends Component {
...
     }
 
-    this.props.onSubmit(this.state.mode, null, plantProject);
+    let formvalue =
+      this.state.mode == 'single-tree'
+        ? this.state.formValueSingle
+        : this.state.formValueMultiple;
+
+    this.props.onSubmit(this.state.mode, formvalue, plantProject);
     event.preventDefault();
```
mahmed0715: @rabiul042 React never allows you to directly set a params in this.state.
You have to use this.setState() to update any state params...
Please check react docs

rabiul042:
Okay, Thanks

app/containers/RegisterTrees/index.js
```diff
@@ -27,12 +27,18 @@ class RegisterTreesContainer extends PureComponent {
...
   onSubmit = (mode, registerTreeForm, plantProject) => {
-    registerTreeForm =
+    /*registerTreeForm =
       registerTreeForm || this.refs.registerTrees.refs.registerTreeForm;
     // console.log(registerTreeForm.validate());
     let value = registerTreeForm.getValue();
     value = Object.assign({}, value);
+    value = mergeContributionImages(value);*/
+
+    let result = this.refs.registerTrees.refs.registerTreeForm.validate();
```
mahmed0715: Please run 
```npm run lint:errors```
and fix eslint errors.

**Comment by mahmed0715** [comment]
@norbertschuler please  check this PR, is this the correct fix?

File: app/server/formSchemas/registerTrees.js [hunk 3]
```diff
@@ -130,7 +130,7 @@ export default {
         propertyOrder: 8
       }
     },
-    required: ['treeCount', 'plantDate', 'geoLocation'],
```
norbertschuler: Who has made the decision that geolocation is not a required input value for tree registration any more?

norbertschuler: Btw. (not part of this patch) if one does not enter a tree type name then the resulting entries at my forest shows "null" for it which does not look very good, but is another error:
Bildschirmfoto 2019-11-13 um 16 30 22
Also deleting a tree from my forest still does not update the client which again is another error #1100

mahmed0715:
### @norbertschuler the api is not asking for geolocation anymore. I checked that.

norbertschuler:
> @norbertschuler the api is not asking for geolocation anymore. I checked that. This field is nit required anymore.

Yes, I saw that. My point was just, do we really want to allow registrations of trees without location? @sagararyal

sagararyal:
@norbertschuler @mahmed0715 

Geolocation should always be required. Else this will create issues in future.

app/server/formSchemas/registerTrees.js
norbertschuler: So as Sagar said `geoLocation` is an mandatory field, let's keep it. If the backend does not throw an error if it’s missing then that’s a backend issue to be solved.

mahmed0715: Changed geolocation as mandatory.

mahmed0715:
@norbertschuler I think its good now to be merged. @rabiul042 as first timer.

PR #1635, opened 2019-11-14
Feature/currency by location : replacing old PR

mahmed0715:
This is updated PR, which uses old currency rates api.


mahmed0715:
replacing https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1572

mahmed0715:
@norbertschuler this is a new PR which uses old currency api. No change in calculation @sagararyal

app/actions/fetchLocation.js
```diff
@@ -0,0 +1,52 @@
...
+          console.error(error);
+        });
+    }
+  };
+}
+
+export function fetchCurrencies() {
+  return dispatch => {
+    getExternalRequest({
+      endPoint: `https://openexchangerates.org/api/latest.json?app_id=${
```
norbertschuler: I don't understand this. In #1572 we changed using this endPoint directly in the client towards using a new API call `getRequest('public_currencies_get', { version: 'v1.3' })` developed by Jorgo to limit the number of API calls (as if every client calls it we would produce a lot more API requests as if our backend makes the request and caches it for all clients). I saw that in the backend code base - is that right @jmiridis? Now we don't want to tunnel this request through our backend any more again, do we?

app/config/index.js.dist
```diff
@@ -22,5 +22,7 @@ export const context = {
   },
   ios:{
     appId: '1444740626'
-  }
+  },
+  locationApikKey: '1dfab408903c332b660f5e7cb90da06a',
+  currencyApiKey: 'd581a5406d694920aa493b9143a9e257'
```
norbertschuler: If we move the currency API to the backend again, a API key does not need to be in the client as well.

mahmed0715: true. removed.

app/actions/fetchLocation.js
mahmed0715: sill we are using fetchcurrencies from actions/currencies.js 
now removed this one.

mahmed0715:
@jmiridis please update the currency rates frequently, I see its 18 days old.

mahmed0715:
@norbertschuler its clean now. please step forward.

mahmed0715:
Lets merge this asap if good. because this will fix an issue reported by user https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1640. @norbertschuler

PR #1650, opened 2019-11-21
React.lazy implement  - [UPDATE]

haideralishah:
According to @Chris Sattinger's suggestion again I implement React.lazy.

### DIFF
--- a/app/components/App/TreeCounter.js
+++ b/app/components/App/TreeCounter.js
@@ -1,38 +1,66 @@
 /* eslint-disable no-underscore-dangle */
 // Library imports
-import React, { Component } from 'react';
+import React, { Component, lazy, Suspense } from 'react';
 import { Route, Redirect, Switch } from 'react-router-dom';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { NotificationContainer } from 'react-notifications';
 import PropTypes from 'prop-types';
+import LoadingIndicator from '../../components/Common/LoadingIndicator';
 
 // Components imports
-import SelectPlantProjectContainer from '../../containers/SelectPlantProject';
-import GiftTreesContainer from '../../containers/GiftTrees';
-import TargetContainer from '../../containers/TargetContainer';
-import RegisterTreesContainer from '../../containers/RegisterTrees';
-import HeaderContainer from '../../containers/HeaderContainer';
-import UserContributionsContainer from '../../containers/UserContributions';
-import EditUserContributionContainer from '../../containers/EditUserContribution';
-import SignUpContainer from '../../containers/Authentication/SignUpContainer';
-import LoginContainer from '../../containers/Authentication/LoginContainer';
-import ForgotPasswordContainer from '../../containers/Authentication/ForgotPasswordContainer';
-import ResetPasswordContainer from '../../containers/Authentication/ResetPasswordContainer';
-import EmailSentContainer from '../../containers/Authentication/EmailSentContainer';
-import SignupSuccessPage from '../Authentication/SignupSuccessPage';
-import BrowserRouter from '../Common/BrowserRouter';
-import SideMenuContainer from '../../containers/Menu/SideMenuContainer';
-import FAQContainer from '../../containers/FAQ';
-import PledgeContainer from '../../containers/Pledge';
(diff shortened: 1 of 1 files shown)


app/components/App/TreeCounter.js
```diff
@@ -1,66 +1,119 @@
...
-import ChallengeContainer from '../../containers/Challenge/createChallenge';
-import RedirectedPublicDenyEmail from '../../containers/Challenge/RedirectedPublicDenyEmail';
-import RedirectedPrivateAcceptEmail from '../../containers/Challenge/RedirectedPrivateAcceptEmail';
+const PrivacyContainer = lazy(() => import('../../containers/Privacy'));
+const ImprintContainer = lazy(() => import('../../containers/Imprint'));
+const AppPaymentContainer = lazy(() => import('../../containers/AppPayment'));
+const BodyErrorBoundary = lazy(() =>
+  import('../ErrorBoundry/bodyErrorBoundry')
+);
+const PageNotFound = lazy(() => import('../ErrorBoundry/404'));
```
crucialfelix: Also no lazy on the 404 display. It should be instantly available.

A page that is this small is not so useful to lazy load. It is just one display component.

app/components/App/TreeCounter.js
```diff
@@ -335,16 +388,24 @@ class TreeCounter extends Component {
...
+            <div style={{ margin: '0 auto' }}>
+              <LoadingIndicator />
+            </div>
+          )}
+        >
+          <BrowserRouter history={history}>
+            <div className="app-container">
+              <ProgressModal />
+              <HeaderContainer />
+              <Route component={SideMenuContainer} />
```
crucialfelix: Does it work to put the Suspense right here around `{this._appRoutes}` ? 

Because you would want the header and footer not to disappear while a page is loading. It should just be the middle page content area that shows loading...

app/components/App/TreeCounter.js
```diff
@@ -1,66 +1,119 @@
...
-import Footer from '../Footer';
+const SelectPlantProjectContainer = lazy(() =>
+  import('../../containers/SelectPlantProject')
+);
+const GiftTreesContainer = lazy(() => import('../../containers/GiftTrees'));
+const TargetContainer = lazy(() => import('../../containers/TargetContainer'));
+const RegisterTreesContainer = lazy(() =>
+  import('../../containers/RegisterTrees')
+);
+const HeaderContainer = lazy(() => import('../../containers/HeaderContainer'));
```
crucialfelix: The Header is not a page: it is always there at the top of the application. So no lazy.

norbertschuler:
I merged the webpack 4 upgrade #1651  to develop and rebase this pull request on develop (please check out your local branch newly).

app/components/App/TreeCounter.js
```diff
@@ -335,16 +388,24 @@ class TreeCounter extends Component {
...
-            <ProgressModal />
-            <HeaderContainer />
-            <Route component={SideMenuContainer} />
-            {this._appRoutes}
-            <Footer />
-          </div>
-        </BrowserRouter>
-        <NotificationContainer />
+        <Suspense
+          fallback={() => (
```
norbertschuler: Got this in the browser console:
```
Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it.
    in Suspense (created by TreeCounter)
    in div (created by TreeCounter)
    in TreeCounter (created by ConnectFunction)
    in ConnectFunction (created by App)
    in GlobalErrorBoundary (created by App)
    in Provider (created by App)
    in App
```

app/components/App/TreeCounter.js
norbertschuler: BrowserRouter and Footer also shouldn't be lazy for this to work. But then it works well.

norbertschuler:
This patch will probably decrease the main bundle size from 2.6MB to 1.5MB and the size of loading the full home page from 3.8MB to 3.1MB.

app/components/App/TreeCounter.js
crucialfelix: I think this (as it was merged) makes the SideMenu disappear every time you click on a new unloaded page.

PR #1645, opened 2019-11-19
Code splitting plus webpack 4

haideralishah:
`React.lazy()` Implement the whole project components Mobile and Web both.

This also upgrades to webpack 4.

PR #1651, opened 2019-11-21
upgrade webpack to version 4

### LINKED ISSUE #1320: Update to webpack 4
It caches compiled results. Build time and hot reload is much faster.

https://webpack.js.org/migrate/4/

Many of the loaders need to be upgraded and checked as well.

norbertschuler:
I took the part which upgrade to webpack 4 from #1645 as a separate pull requests here.

Fix #1320


web/config/webpack.prod.config.js
```diff
@@ -50,15 +54,6 @@ module.exports = webpackMerge(commonConfig, {
...
         }
       }
-    }),
-    new webpack.NoEmitOnErrorsPlugin(),
-    new webpack.optimize.UglifyJsPlugin({
-      mangle: {
-        keep_fnames: true
-      }
-    }),
-    new webpack.DefinePlugin({
```
crucialfelix: Is this taken care of automatically by webpack now? Is `process.env.NODE_ENV` replaced in the compiled source?

norbertschuler: According to https://webpack.js.org/migrate/4/ : "These plugins can be removed from configuration as they are default in production mode"

PR #1657, opened 2019-11-23
Feature/image optimization

haideralishah:
Webpack: image Optimization is done.


norbertschuler:
@haideralishah have you seen https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1656? Is this an addon or have you done this independent of my PR?

web/config/webpack.dev.config.js
```diff
@@ -2,6 +2,7 @@ const webpack = require('webpack');
 const webpackMerge = require('webpack-merge');
 const commonConfig = require('./webpack.common.config.js')();
 const path = require('path');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
```
norbertschuler: This is already imported and defined in `webpack.common.config.js` resulting into an error for me when running `npm start` as following:

```
WARNING in Conflict: Multiple assets emit different content to the same filename 3.css.map
WARNING in Conflict: Multiple assets emit different content to the same filename bundle.css.map

```

web/config/webpack.dev.config.js
```diff
@@ -39,9 +40,32 @@ module.exports = webpackMerge(commonConfig, {
...
+              options: {
+                bypassOnDebug: true, // webpack@1.x
+                disable: process.env.node_env !== 'production' // webpack@2.x and newer
+              }
+            }
+          ]
+        }
+      ]
+    })
+  ]
```
norbertschuler: Why should we just define this plugin in our dev configuration file and not for production?

package.json
```diff
@@ -57,7 +57,7 @@
     "react-i18next": "^7.6.1",
     "react-images": "^0.5.17",
     "react-modal": "^3.11.1",
-    "react-native": "^0.59.8",
+    "react-native": "0.59.8",
```
norbertschuler: Why have you fixed the react-version. Is there a known compatibility problem with higher versions?

haideralishah: when I run `npm install` `react-native` update to 0.60.4 which is the latest version of `react-native`  so any things are deprecated in the latest version like 'ListView'. so I fixed the react-native version for temporary.

norbertschuler: O.k., that shouldn't happen of course. I wonder why it does not upgrade for me running `npm install`.

norbertschuler:
Closed in favor of #1656 - @haideralishah thank you very much for your work. I am sorry, I did the same at the same time and therefore we had duplicated work. I used this patch to compare it to mine to check that I have not forgotten anything so your work helped our project to become better.

PR #1658, opened 2019-11-23
fix formatnumber.replace is undefined issue

### LINKED ISSUE #1588: Check formatNumber function not always returning a string

**TypeError** in **MainActivity**
TypeError: TypeError: TypeError: (0,T.formatNumber)(1,null,l).replace is not a function. (In '(0,T.formatNumber)(1,null,l).replace(/[\d.,]/g,'')', '(0,T.formatNumber)(1,null,l).replace' is undefined)

This error is located at https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/33311ed8d32d39cb81fe42738e83d04328f9a630/app/components/Currency/TreeCountSelector.js#L163 and at https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/33311ed8d32d39cb81fe42738e83d04328f9a630/app/components/Currency/TreeCountSelector.native.js#L239 as the code relies on a string to be returned from the function formatNumber https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/33311ed8d32d39cb81fe42738e83d04328f9a630/app/utils/utils.js#L47

[View on Bugsnag](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5dbd81cefa513b00183b45fc?event_id=5dbd81ce00525d5915880000&i=gh&m=ci)

## Stacktrace

    app/components/Currency/TreeCountSelector.native.js:240 - value
    app/components/Currency/TreeCountCurrencySelector.native.js:69 - value
    app/components/Currency/TreeCountCurrencySelector.native.js:40 - value
    app/components/Currency/CurrencySelector.native.js:54 - onChangeText

[View full stacktrace](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5dbd81cefa513b00183b45fc?event_id=5dbd81ce00525d5915880000&i=gh&m=ci)

mahmed0715:
Fix #1588 
The /currencies api had an issue with data like we get "AWG" currency in currency names array but thats not present on the rates, thus having issue NAN while we change currency to AWG(just an example)
@jmiridis need to check this issue.

### DIFF
--- a/app/components/Currency/TreeCountSelector.js
+++ b/app/components/Currency/TreeCountSelector.js
@@ -92,9 +92,17 @@ class TreeCountSelector extends React.Component {
         : newState.variableAmount
     });
   }
-
+  getFormattedNumber(treeCount, symbol) {
+    const { currency, treeCountToAmount } = this.props;
+    try {
+      const data = formatNumber(treeCountToAmount(treeCount), null, currency);
+      return symbol ? data.replace(/[\d.,]/g, '') : data;
+    } catch (err) {
+      return symbol ? currency : treeCountToAmount(treeCount) + ' ' + currency;
+    }
+  }
   render() {
-    const { treeCountOptions, currency, treeCountToAmount } = this.props;
+    const { treeCountOptions } = this.props;
 
     return (
       <div className={'treecount-container'}>
@@ -118,7 +126,7 @@ class TreeCountSelector extends React.Component {
               </label>
               <span className="price-conversion__equal">=</span>
               <span className="price-conversion__amount">
-                {formatNumber(treeCountToAmount(treeCount), null, currency)}
+                {this.getFormattedNumber(treeCount)}
               </span>
             </div>
           );
@@ -160,7 +168,7 @@ class TreeCountSelector extends React.Component {
                 this.handleVariableAmountChange(evt.target.value)
               }
             />{' '}
-            {formatNumber(1, null, currency).replace(/[\d.,]/g, '')}
+            {this.getFormattedNumber(1, true)}
           </span>
         </div>
       </div>
--- a/app/components/Currency/TreeCountSelector.native.js
+++ b/app/components/Currency/TreeCountSelector.native.js
@@ -88,9 +88,19 @@ class TreeCountSelector extends React.Component {
         : newState.variableAmount
     });
   }
+  getFormattedNumber(treeCount, symbol) {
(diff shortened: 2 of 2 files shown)


mahmed0715:
Jorgo and me had a talk and fixed the issue. @norbertschuler @harshvitra please check if that issue fixed alrady in this branch

norbertschuler:
I have no idea on how to test this. @mahmed0715 How can the error be tested best? And will it still occur after @jmiridis removed some currencies from the backend with https://github.com/Plant-for-the-Planet-org/treecounter-platform/commit/01aa12950fb697c2c2454dd0e46ee562be06de10?

mahmed0715:
I see somehow on a very first run(I am not sure why) the default currency is not set(missing on the currency selector left side). Then I see the .replace issue, but then I never be able to reproduce it.
I think in this PR how I manged the formatnumber method call will fix future issue like this..

I think if you see this PR is working good, then its okay to merge(at least the .replace issue will never come up) @norbertschuler

app/components/Currency/TreeCountSelector.js
```diff
@@ -92,9 +92,17 @@ class TreeCountSelector extends React.Component {
         : newState.variableAmount
     });
   }
-
+  getFormattedNumber(treeCount, symbol) {
+    const { currency, treeCountToAmount } = this.props;
+    try {
+      const data = formatNumber(treeCountToAmount(treeCount), null, currency);
+      return symbol ? data.replace(/[\d.,]/g, '') : data;
```
norbertschuler: As the function `formatNumber` just returns the first parameter (in this case `treeCountToAmount(treeCount)`) when it fails to format it, I assume the value of `treeCountToAmount(treeCount)` was undefined. So there is definitely a use case where something is missing (e.g. rates or selected currency) which causes this error and should further be investigated.

If `treeCountToAmount(treeCount)` is undefined returning `treeCountToAmount(treeCount) + ' ' + currency` won't help us very much. It's still an error for the user. We just prevent the app from crashing.

mahmed0715: yes, thats my intention for this PR to just prevent the crash, I think need more investigation is required to findout exact issue.

mahmed0715:
True @norbertschuler 
 we need to find out the exact issue, merging this and creating another ticket so that someone can work on that..

PR #1659, opened 2019-11-24
Update Errormessage.js

mahmed0715:
Fix eslint error

### DIFF
--- a/app/components/Common/Text/Errormessage.js
+++ b/app/components/Common/Text/Errormessage.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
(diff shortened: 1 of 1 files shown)


app/components/Common/Text/Errormessage.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
+// eslint-disable-next-line no-unused-vars
 const Errormessage = ({ children, strong }) => (
```
norbertschuler: Is `strong` maybe unused?

mahmed0715: yes, for now it is no being used.

Issue #1615, opened 2019-11-09
Add verified or self-reported status to Survival Rate


sagararyal:
As we allow projects to report the survival rate themselves, it is important to convey the users a message whether the information is correct or not. 

As we have started to see projects reporting 100% survival rates, 

The goal is to show which are self-reported, and which are certified by trusted organisations.

TPOs can submit the reports to have the certification status added.

In project cards

If a project is certified:
<img width="577" alt="Screen Shot 2019-11-09 at 11 05 31 AM" src="https://user-images.githubusercontent.com/5401575/68526881-ebc97600-02e0-11ea-8172-cbec51c4eafc.png">

if it is not certified:

<img width="595" alt="Screen Shot 2019-11-09 at 11 05 40 AM" src="https://user-images.githubusercontent.com/5401575/68526889-fe43af80-02e0-11ea-849f-629b07248e7b.png">

Opening the card 

<img width="747" alt="Screen Shot 2019-11-09 at 11 25 34 AM" src="https://user-images.githubusercontent.com/5401575/68527112-a8bcd200-02e3-11ea-8f0a-67440dcd767e.png">

mahmed0715:
@sagararyal I am doing this, currently we are getting 3 related params in the projects data
```
survivalRateStatus:"self-reported",
isVerified:true,
isCertified:false
```
so to confirm, depending on these params we will change the view, right?
The opening the card means details view of the project will show verified , Self reported status?(will not show in the card/list view)

sagararyal:
@mahmed0715 

this does not concern `isverified` or `iscertified` flags

display them based on:
```
survivalRateStatus:"self-reported",
survivalRateStatus:"verified",
```

mahmed0715:
This can be closed @sagararyal ?

norbertschuler:
> This can be closed @sagararyal ?

I thought we keep this issue as we maybe also want to add this to the web client and maybe we want to implement the alternative view when looking at the details of a project?

PR #1660, opened 2019-11-24
Added survival rate status leaf to card

mahmed0715:
Some part of https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1615 has been implemented with this.
@sagararyal please check


app/components/PlantProjects/PlantProjectSnippet.native.js
```diff
@@ -200,9 +214,33 @@ class PlantProjectSnippet extends PureComponent {
...
+                            source={survivalRateLeaf}
+                            style={{
+                              width: 13,
+                              height: 13
+                            }}
+                          />
+                        </Text>
+                      ) : null}
+                    </View>
+                  </View>
```
norbertschuler: Text color and text size differ from the line below and above this new text - maybe they should be adapted to the same style?

Bildschirmfoto 2019-11-26 um 10 17 04

Doesn't the style `styles.survivalText` apply to the text component any more?

app/components/PlantProjects/PlantProjectSnippet.native.js
```diff
@@ -97,6 +105,12 @@ class PlantProjectSnippet extends PureComponent {
         deducibleText1 += ', ';
       }
     }
+    const survivalRateLeaf =
+      survivalRateStatus == 'verified'
```
norbertschuler: There does not seem to be an example on `staging` or `test` backend of a `verified` project, but I assume this will work :-) Is the key `verified` defined for our API?

app/components/PlantProjects/PlantProjectSnippet.native.js
mahmed0715: hmm, this style should be applied, I will see why i was not seeing the change.

app/components/PlantProjects/PlantProjectSnippet.native.js
mahmed0715: yes, verified/self-reported are the keys

sagararyal:
@mahmed0715 I built this on both staging and test, but I don't see the leaf in any projects.

norbertschuler:
> @mahmed0715 I built this on both staging and test, but I don't see the leaf in any projects.

It just shows them if the API call /loadTpos returns the property `survivalRateStatus` which worked for me on `staging` early this morning.

sagararyal:
Still the part inside is left, but we can do it in another task, which is more design focused.
Approval from my side.

app/components/PlantProjects/PlantProjectSnippet.native.js
mahmed0715: updated code @norbertschuler pls see

mahmed0715: but how come I diddnt see the change .

mahmed0715:
@norbertschuler pls check

app/components/PlantProjects/PlantProjectSnippet.native.js
norbertschuler: I tested it with iOS simulator meaning the Safari browser engine.

mahmed0715:
merging....
But We discussed about the project details page, that another context, we will improve web and do in other way.

PR #1668, opened 2019-11-26
Support tree context by slug

mahmed0715:
`/gift-trees/<slug>` now supported and if slug found the givee is set and user is being taken to project selection..
@sagararyal please check on web.
What about App?

### DIFF
--- a/android/.project
+++ b/android/.project
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-	<name>Plant</name>
+	<name>TreecounterApp</name>
 	<comment>Project android created by Buildship.</comment>
 	<projects>
 	</projects>
--- a/app/components/App/TreeCounter.js
+++ b/app/components/App/TreeCounter.js
@@ -331,12 +331,12 @@ class TreeCounter extends Component {
             />
             {/*<Route path="/payment/project/:projectId" component={PaymentDonation}/>*/}
             <Route
-              path={getLocalRoute('app_giftTrees') + '/:slug?'}
+              path={getLocalRoute('app_giftTrees')}
               component={GiftTreesContainer}
             />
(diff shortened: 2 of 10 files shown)


mahmed0715:
@sagararyal added /support/slug alias but we need to update the url on support button. will do that.
@jmiridis need to add this url in /server/routes/fos_js_routes.json 
I have added for test purpose. Is there a way we dont need to ask @jmiridis  everytime I  need a change in that file?

mahmed0715:
@sagararyal we removed gift-trees/slug and now added /support/slug that shows as Support trees to <User name> and we do a donation with commuityTreecounter.
Please test

norbertschuler:
Is there any issue related to this pull request where I could see what the purpose of this change is about?

mahmed0715:
@norbertschuler this is part of contextual donation , we are doing part by part, like this is support, we do project/slug we do other context ..

mahmed0715:
throw an error for wrong slug
and allow user to donate

norbertschuler:
> TODO: throw an error for wrong slug and allow user to donate
Does this mean this PR is not yet ready to be merged?

android/.project
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-	<name>Plant</name>
+	<name>TreecounterApp</name>
```
norbertschuler: This does not change the name of the app (defined at https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/81aee5bd5c83692a326d47f1ecf3b71b92317eb3/android/app/src/main/AndroidManifest.xml#L12) rather than just the Android project name.

sagararyal: @norbertschuler I saw this as well, but we'll do a cleanup of these things on a separate PR.

PR #1673, opened 2019-11-28
update react native packages

### LINKED ISSUE #1327: Update react-native to latest version 0.60.x
We are currently using version 0.59.8 and therefore already need patches like the postinstall script in `package.json` to make the Android build possible.

To better support the latest versions of Android and iOS we should upgrade to react-native 0.60.x (0.61.x is still a release candidate). Unfortunately this upgrade might be quite painful if the Android and iOS project structure has changed and we would have to adapt our individual changes to the Android and iOS project again if they get lost by upgrading (I am not sure about that).

Also while doing so we could upgrade other react-native node modules which are linked in the iOS or Android project:
- react-native-linear-gradient
- bugsnag-react-native
- react-native-device-info
- react-native-image-picker
- react-native-i18n
- react-native-svg
- react-native-youtube
- ...

A helper tool proposed by @harshvitra https://react-native-community.github.io/upgrade-helper/?from=0.59.10&to=0.61.4

Read more about what you should be aware when doing the upgrade at https://facebook.github.io/react-native/blog/2019/07/03/version-60

sanjayradadiya:
Fixes #1327

Changes in this pull request:
- Update react-native version
- Fix ios pod dependencies 
- Fix android dependencies
- Quick fix on motivation and login captcha

Compare with https://react-native-community.github.io/upgrade-helper/?from=0.59.8&to=0.61.4


.eslintrc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  root: true,
+  extends: '@react-native-community'
```
norbertschuler: Actually we better do not extends this rule set. I tried to add it to our `.eslintrc.json`, but then our own rules about using single quotes got overwritten.

app/components/Authentication/Login/index.native.js
```diff
@@ -69,13 +52,6 @@ export default class Login extends Component {
         resetScrollToCoords={{ x: 0, y: 0 }}
         scrollEnabled
       >
-        <ReCaptchaV3
```
norbertschuler: Why do you have deleted ReCaptchaV3? Wasn't is working with react-native 0.6x any more?

norbertschuler: see #1690

norbertschuler:
closed in favor of #1690

Issue #1661, opened 2019-11-25
Swipe top to bottom to refresh management on app


mahmed0715:
From https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/932 we wanted to implement a swipe to refresh mechanism on our app where we reload data on a certain Scene on Swipe top to bottom.

mahmed0715:
I am working on it.

PR #1676, opened 2019-11-28
Refresh Control

### LINKED ISSUE #1608: Integrate RefreshControl for Competitions v1.3
https://facebook.github.io/react-native/docs/refreshcontrol

### LINKED ISSUE #1684: Competition v1.2
- [ ] Currently All and Archived use same API Call, it seems this does not work in production.
Jorgo will provide api for archived

harshvitra:
Fixes #1608
Fixes #1684

Changes in this pull request:
- Added Refresh Control

Test - Scroll on these pages
- Featured Competitions
- All Competitions
- Mine 
- Archived


mahmed0715:
is this something that is part of https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1661 ?

sagararyal:
@mahmed0715 seems to be.

harshvitra:
Yes, as of now I have implemented in Competitions. 

Will implement this in Pledge Events too, as it updates frequently and if doesn't user can try updating it.
Projects don't update that frequently so we can do that later too

harshvitra:
Also fixes #1684

small API update needed - 
Right now both 
`/competitionsMine and /competitions/archived`

APIs have the same **id - 3**, so they replace the content of each other

harshvitra:
> APIs have the same **id - 3**, so they replace the content of each other

All API has id 1
Featured API has id 2
Mine and Archived API has id 3

So in the pager object only one can exist at one point of time

norbertschuler:
- [ ] Testing with the staging backend, refreshing **favorites** and **all** tabs works, but refreshing **mine** or **archived** competitions never removed the refreshing spinner again (even if you switch the tabs) - only pulling down again the screen to do another refresh removes this state (but then doesn't do another API call to refresh the page):

Simulator Screen Shot - iPhone 11 - 2019-12-06 at 10 00 39
Simulator Screen Shot - iPhone 11 - 2019-12-06 at 10 00 28

sagararyal:
In my case, i get error, logging in.

all other branch works, (testing with production data)

Screen Shot 2019-12-08 at 12 56 25 PM

sanjayradadiya:
I have tested this on IOS and here is testing video https://www.loom.com/share/c20c490845174a50a208f8e6e7513a12

some it work's sometimes not
if you pull screen till half of the screen then it will create an issue
even on Featured tab, it's the same issue

harshvitra:
Now the issue can be described as - 

Scrolling half way in iOS on any Tab causes infinite refreshing

@sanjay I'm sending a solution on Slack please test on iOS

sanjayradadiya:
> Now the issue can be described as -
> 
> Scrolling half way in iOS on any Tab causes infinite refreshing
> 
> @sanjay I'm sending a solution on Slack please test on iOS

the solution you provided is not related to this issue 

I have fixed the issue on IOS here is video https://www.loom.com/share/852c450124f642518a86c27bfc63874f

Issue #1666, opened 2019-11-26
label.by_a_name got renamed and changed


norbertschuler:
**Describe the bug**
The translation resource `label.by_a_name` got renamed and its content changed from `by` so it's now missing in our source code.
https://github.com/Plant-for-the-Planet-org/treecounter-app/commit/df0a74738e004e87f200af6254c0d62a8e5e5b9c#diff-6a40449dcc2e1c6c221d85eea53a6592

-> We have to add it again.

**Is it on a Production or Development Build?**
develop

**To Reproduce**
1. Go to https://test.trilliontreecampaign.org/donate-trees

**Screenshots**
<img width="775" alt="Bildschirmfoto 2019-11-26 um 12 54 45" src="https://user-images.githubusercontent.com/1532418/69628846-2efb4700-104c-11ea-893c-9777b91094b4.png">

**Desktop (please complete the following information):**
- OS: macOS
 - Browser latest Chrome
 - Plant-for-the-Planet App Version 1.1.1

Issue #1644, opened 2019-11-19
label.donate is defined twice


norbertschuler:
**Describe the bug**
The resource `label.donate` is defined twice with different translations.

<img width="286" alt="Bildschirmfoto 2019-11-19 um 11 03 24" src="https://user-images.githubusercontent.com/1532418/69159516-57cd8a80-0ae8-11ea-9c9f-c94d591e7940.png">

We will have to check all `label.donate` usages and decide at which place “donate” is wanted and at which place “projects” is wanted and change the places for “projects” e.g. by renaming one of these resources to `label.projects`.

**Is it on a Production or Development Build?**
development

mahmed0715:
on it

norbertschuler:
@mahmed0715 maybe you can do it together with #1666

PR #1682, opened 2019-11-29
label.by_a_name rename and label.projects vs. label.donate fix

rabiul042:
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1666
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1644


app/components/PlantProjects/PlantProjectFull.js
```diff
@@ -143,7 +143,7 @@ class PlantProjectFull extends React.Component {
             <div className="row">
               <div className="teaser__tpoHeading">
                 <a onClick={() => this.updateRoute(teaserProps.tpoSlug)}>
-                  {i18n.t('label.by_a_name') + ' ' + teaserProps.tpoName}
+                  {i18n.t('label.comp_by_name') + ' ' + teaserProps.tpoName}
```
norbertschuler: Using the resource `comp_by_name` without setting its both parameter append a space before and after the resource, e.g. ` by `.

mahmed0715: true @norbertschuler 
@rabiul042 please create a new resource that is only for the text 'by'

norbertschuler: no, I already changed it to use at least one placeholder from `comp_by_name` which also works well as just has a space in front which does not matter in HTML.

norbertschuler:
> There is a use case where @sagararyal wanted to show Screen heading as 'Projects" for the donate trees, thus I needed to change that.
> @sagararyal please confirm.

O.k. sorry, then we have to change the used resource from `label.donate` to `label.projects` in the code. Do not change again the content of the resources.

norbertschuler:
@mahmed0715 please merge if you think everything is done.

mahmed0715:
Thanks

PR #1686, opened 2019-11-30
competetion 404 solved

### LINKED ISSUE #1626: shared link from competition results into 404 on web client
A shared linked for a competition from the app results into 404 error: https://www.trilliontreecampaign.org/competition/55 if opened with mobile Chrome browser on iOS. 

Opening it from the mail app correctly opens the deep link in the app. 

Maybe we later on want to support our web client not to show 404 errors for some deep links which could not be opened with the app on a phone.

All other currently used share links are also supported URI for the web client.

rabiul042:
Fixes #1626

### DIFF
--- a/app/components/App/TreeCounter.js
+++ b/app/components/App/TreeCounter.js
@@ -73,6 +73,9 @@ import { getLocalRoute } from '../../actions/apiRouting';
 const SuccessfullyActivatedAccount = lazy(() =>
   import('../../containers/Authentication/SuccessfullActivatedContainer')
 );
+const CompetitionContainer = lazy(() =>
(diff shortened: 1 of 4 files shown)


mahmed0715:
LGTM

norbertschuler:
This results into showing the users the following screen from https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/accd1f4cf0cc4f79e457f4df5dc90c50ff522a23/app/components/Competition/index.js#L7

Bildschirmfoto 2019-11-30 um 14 43 10

Do we want to add at least one nicer (and translated) text?

mahmed0715:
True, will do. @norbertschuler

Issue #1670, opened 2019-11-27
Donate Trees : release candidate v1.2


mahmed0715:
- [x] In Single Projects Page, the price doesn't work property, so implement a floating button.
- [x] In project list view target, comma is missing on number less than 100k eg (100,000)
- [x] Fix Donation view, when support is being clicked, Instead of bring users to Donate trees without option to go back

mahmed0715:
@sagararyal `In Single Projects Page, the price doesn't work property, so implement a floating button.` need explanation please

sagararyal:
@mahmed0715 
the price per tree needs to be tapped twice to go to the donate screen,

instead of fixing that, which will be removed in future updates, you could have the Donate button, which is at the bottom of the screen be a floating button, like it is in Edit Competition and other places.

this will be removed in future as well.

mahmed0715:
on it.

norbertschuler:
@sagararyal check out this floating donate button @mahmed0715 has build:

![Simulator Screen Shot - iPhone 11 - 2019-12-06 at 15 35 52](https://user-images.githubusercontent.com/1532418/70330589-266df200-183e-11ea-9107-6262efafce1b.png)

mahmed0715:
@norbertschuler the background texts are scrollable, right?
 so all the contents behind the button are visible if scrolled. right?

PR #1693, opened 2019-12-03
Donation v1.2: release candidate

mahmed0715:
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1670

Repeated from issue:
- [x] In Single Projects Page, the price doesn't work property, so implement a floating button.
- [x] In project list view target, comma is missing on number less than 100k eg (100,000)
- [x] Fix Donation view, when support is being clicked, Instead of bring users to Donate trees without option to go back


mahmed0715:
1. solved : price is now good cliable, Float the donate button

mahmed0715:
2. solved. format target (i18n works)

mahmed0715:
@sagararyal the back button from support button works now.

Note for developer: Added inner route app_supportTrees, so next time when we need to redirect donate from support button we need to use this inner route.

app/components/PublicTreeCounter/PublicTreecounter.native.js
```diff
@@ -54,7 +54,7 @@ class PublicTreeCounter extends React.Component {
 
   onRegisterSupporter() {
     this.props.supportTreecounterAction(this.props.treecounter);
-    updateRoute('app_donateTrees', this.props.navigation, 0, {
+    updateRoute('app_supportTrees', this.props.navigation, 44, {
```
norbertschuler: Just as I am curious - how did you decide about using the number `44` as the id parameter?

app/styles/selectplantproject/selectplantproject-full.native.js
```diff
@@ -53,6 +53,12 @@ export default EStyleSheet.create({
     flexShrink: 0,
     padding: 16
   },
+  floatingButtonContainer: {
+    padding: 16,
+    position: 'absolute',
+    bottom: 35,
+    right: 0
+  },
```
norbertschuler: I am not a designer, but I would prefer a rounded button as @harshvitra used them for competitions or pledge events. Maybe he can share his styles with you. Also an short action button as `(+)` is commonly positioned at the bottom right edge of the screen while a bigger button with text on it as you designed it is more often centered at the bottom of the screen.

app/components/Competition/CompetitionParticipant.native.js
```diff
@@ -22,7 +22,7 @@ class CompetitionParticipant extends React.Component {
     };
     console.log('support button pressed');
     this.props.supportTreecounterAction(supportObject);
-    updateRoute('app_donateTrees', this.props.navigation, 0, {
+    updateRoute('app_supportTrees', this.props.navigation, 55, {
```
norbertschuler: Just as I am curious - how did you decide about using the number 55 as the id parameter? What will happen if we accidentally use the same number twice?

app/components/PublicTreeCounter/PublicTreecounter.native.js
mahmed0715: The code looks for only 0 value, so without 0 anything will work...

app/styles/selectplantproject/selectplantproject-full.native.js
mahmed0715: For this case we need donate text over there, I can center it, but @sagararyal suggested a button like we have on competition like the + button.. we can not use + for this case...
@sagararyal can you please check the current button

app/components/Competition/CompetitionParticipant.native.js
mahmed0715: same as before, the logic checks for only 0 value, so any value other than 0 will work..

app/components/PublicTreeCounter/PublicTreecounter.native.js
norbertschuler: Ok, right, it does not seem to matter for https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/27deef1adee82362e233544a3be7def314ccb315/app/helpers/routerHelper/routerHelper.native.js#L11 I had been looking at the file for the web client where the id is used for a state at https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/27deef1adee82362e233544a3be7def314ccb315/app/helpers/routerHelper/routerHelper.js#L7

app/styles/selectplantproject/selectplantproject-full.native.js
norbertschuler: So maybe at least use rounded corners ;-)

sagararyal: clarification I suggested button with rounded corners like in Edit/ Add Competition pages, not + button.

mahmed0715: implemented new design only for bottom area on this PR

mahmed0715:
@sagararyal please check the bottom area is implemented on this PR, we can merge, 
I will create another PR with the new design.

mahmed0715:
Screenshot_1575868480
latest look

sagararyal:
There is a known issue for iOS, but I am going to merge this and create a new issue for that.

Issue #1685, opened 2019-11-30
PledgeEvents v1.2


norbertschuler:
- [ ] Add validation in the pledge Increment screen
- [ ] Currently it is possible to reduce pledges, api returns error but app does not react.
- [ ] Show all details in the Pledge Increment Screen
(allow user to change name or toggle anonymous status, disable email field)
- [x] Fix headers, remove title, and remove spacing 
- [ ] Refresh the page after pledge is updated (including the total pledged tree count)

harshvitra:
@jmiridis 
`allow user to change name or toggle anonymous status, disable email field`

Need an Edit Pledge API which can allow the user to edit these fields 
1. First Name
2. Last Name
3. Email
4. Tree Count
5. Is Anonymous Status

harshvitra:
Hey @norbertschuler  can you explain these issues 


Is number 1 and 2 the same or different ? 

Fix headers, remove titles, and remove spacing - Need more ideas on this ?
 
Refresh the page - Which page ? In the app or website ? Mostly I’m guessing Pledge Events screen on App.

norbertschuler:
Discussion done in Slack (just to inform any possible reader here).

harshvitra:
**For others to be clear**

- Add validation in the pledge Increment screen
- Currently it is possible to reduce pledges, api returns error but app does not react.

These are listed as 2 different but are the same
-------------------------------------------------------------------------------------------

-  Show all details in the Pledge Increment Screen (allow user to change name or toggle anonymous status, disable email field)

So user can change everything and NOT change email
-------------------------------------------------------------------------------------------

- Fix headers, remove title, and remove spacing ###DONE


Fixing - Removing Search icon, changing color 
Remove title - removing titles and adding them in header
Remove spacing - unnecessary
-------------------------------------------------------------------------------------------

-  Refresh the page after pledge is updated (including the total pledged tree count) 

This still needs to be tested and replicated
Cannot test since I cant create Pledges as of now due to error

harshvitra:
@jmiridis can we fix #1692

we cannot solve other issues until this one is fixed

norbertschuler:
> @jmiridis can we fix #1692
> 
> we cannot solve other issues until this one is fixed

Problem got fixed says @jmiridis.

Issue #1695, opened 2019-12-03
Remove search icon from Pledge Events


harshvitra:
Pledge events currently has search icon, which has no purpose as of now in the Application to be on the pledge events page and create/update pledge page

norbertschuler:
I assume this is fixed as #1697 got merged.

harshvitra:
Can we wait for #1731 to get merged, we can leave it closed since no other page is involved in the flow

Issue #1694, opened 2019-12-03
Add Header Animations to Pledge Events Feature


harshvitra:
https://material.io/components/assets/1d3jb571H-QqU8rMvcvksU1e8-Di0lgu0/topappbars-behavior-scrolling-prominent-to-regular.mp4

norbertschuler:
@harshvitra is this issue related to #1872?

PR #1697, opened 2019-12-03
Fixed One issue of Pledge events v1.2 regarding Headers

harshvitra:
Fixes one issue from #1685 

**### Added 3 Types of Header components**
1. Header with no Animations
2. Header with Title Animation
3. Header with Title and Image Animation

Removed Search bar and unnecessary spacing

### The Headers do not support -
1. Search
2. Share
3. Burger Menu
etc

These shall be added soon

@sagararyal please check the Headers and animations


enroot-mumbai:
Fixes #1685
Fixes #1695
Fixes #1694

norbertschuler:
- [ ] On an iPhone X simulator the header does not work (arrow is not clickable) as the header does not check the top safety area - I assume that's because we don't regard it at https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/69c03b95e6a04b2f582484f94acb01d93f0592a9/app/components/App/TreeCounter.native.js#L27 - but please do not change that as otherwise our current header would not work correctly. Maybe you can add an additional `<SafeAreaView>` for you own header:

Simulator Screen Shot - iPhone 11 - 2019-12-06 at 13 41 56

Simulator Screen Shot - iPhone 11 - 2019-12-06 at 13 42 05

app/components/Header/HeaderNew.native.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { Text, View, TouchableOpacity, Image, BackHandler } from 'react-native';
+import { backArrow } from './../../assets';
+
+export default function HeaderNew(props) {
```
norbertschuler: This component is not used, is it?

harshvitra: Once this PR is merged I'm planning to use this and other 2 components to change all the headers in the app

harshvitra:
> * On an iPhone X simulator the header does not work (arrow is not clickable) as the header does not check the top safety area - I assume that's because we don't regard it at

can you test if the latest commit fixed it for you

norbertschuler:
> can you test if the latest commit fixed it for you

I can earliest test this again at Monday.

sagararyal:
@harshvitra https://facebook.github.io/react-native/docs/safeareaview

norbertschuler:
I made some tests with an Nexus 4 API Level 27 (Andriod 8.1) emulator with these results (first one has highest priority, the other two are just minor design issues):

- [ ] I could not click the button to submit a new pledge as it was not on the visible screen and also not reachable by scrolling:

Screenshot_1575877711

- [ ] some texts in event pledges boxes got cut 

Screenshot_1575877641

- [ ] The pledge event name is too long for the screen, but does not get ellipsed:

Screenshot_1575877720

norbertschuler:
> @harshvitra https://facebook.github.io/react-native/docs/safeareaview

Yes, why can't we solve this with given solution from react-native as a SafeAreaView instead of trying to fix it with an own solution?

norbertschuler:
> > * On an iPhone X simulator the header does not work (arrow is not clickable) as the header does not check the top safety area - I assume that's because we don't regard it at
> 
> can you test if the latest commit fixed it for you

- [ ] I did a new test with my iPhone X simulator with the given fix, but it still does not work:

Simulator Screen Shot - iPhone 11 - 2019-12-09 at 09 07 48

harshvitra:
> Yes, why can't we solve this with a given solution from react-native as a SafeAreaView instead of trying to fix it with an own solution?

I've added SafeAreaView

But cant get iphone simulator to build - 
Screen Shot 2019-12-09 at 11 20 00 PM

norbertschuler:
> I've added SafeAreaView
> But cant get iphone simulator to build 

I am sorry, I can not see the reason for the build failure from your screen shot. Maybe you can try clean your build or rerun the pod install with `pod install` in your ios folder.

norbertschuler:
@harshvitra which SafeAreaView do you use? We have two different components from `react-navigation` and `react-native` used within our sources (don't ask me why ;-)):
Bildschirmfoto 2019-12-09 um 20 25 25

harshvitra:
> I am sorry, I can not see the reason for the build failure from your screen shot. Maybe you can try clean your build or rerun the pod install with `pod install` in your ios folder.

I did this already - 

'fatal error: 'UIKit/UIUserActivity.h' file not found'

This is what I can read in terminal

norbertschuler:
> > I am sorry, I can not see the reason for the build failure from your screen shot. Maybe you can try clean your build or rerun the pod install with `pod install` in your ios folder.
> 
> I did this already -
> 
> 'fatal error: 'UIKit/UIUserActivity.h' file not found'
> 
> This is what I can read in terminal

Probably react-native 0.6x requires at least Xcode 10.x to work and Xcode 9.x is not supported any more: https://github.com/facebook/react-native/issues/25492

norbertschuler:
> Check out some article about the different versions of SafeAreaView: https://medium.com/@maheshnandam/use-safeareaview-from-react-navigation-to-fix-iphone-x-design-issue-ba531a9181d

As we now use v4 of `react-navigation` I think we have case 1 and should only more use `import {SafeAreaView} from 'react-navigation';` without **forceInset** and with a proper **headerStyle**.

sanjayradadiya:
https://www.loom.com/share/7ad1571f4db7402aa067fa101ff39d9b

@harshvitra  you need to change some CSS 
There is lot's of white space

I am not sure it's in design or something we have put that intentionally

PR #1710, opened 2019-12-05
Feature/welcome screens

haideralishah:
Update Dec 17

- [x] Fix issue with OpenSans font in iOS (on App Start)

- [x] Swipe right to left does not work in iOS

- [x] disable top bottom scroll, if height of screen is sufficient to display the content.

---
@haideralishah I reviewed the apk you shared with me.

this PR does nothing.

- [x] remove fat close button

- [x] only make the middle content swipe able

----------------------------------------------
I have Completed Welcomes UI screens
For running a demo just call  WelcomeComponent from  `app/components/Welcome/WelcomeSlider.js` here.
let me know If need any modification or enhancement.


package.json
```diff
@@ -59,6 +59,7 @@
     "react-modal": "^3.11.1",
     "react-native": "^0.59.8",
     "react-native-action-button": "^2.8.5",
+    "react-native-app-intro-slider": "^3.0.0",
```
norbertschuler: I cannot see where this module is used?

haideralishah: I removed this module.

norbertschuler:
This build needs some eslint fixes. Check with `npm run lint:errors`.

harshvitra:
The one ESlint error is temporary 

We don't have the conditions to load the Welcome Screen for the user yet so I've commented appContainer and loaded welcome screens for now 

@haideralishah can you think of the user flow ? 
We need to show these screens only once after user installs

PR #1713, opened 2019-12-06
Added invoice_numer : ttc-donationId to paypal, to track payment source

### LINKED ISSUE #1701: Add donations reference to Paypal payment requests
The information that is currently submitted to Paypal to capture and execute a payment, does not contain any reference to the donation and not even the fact that it was created in TrillionTreeCampaign.

We have to identify what additionals fields could be used to provide a reference to the donation so it does appear in the Paypal backend and people in the office can identify the origin of the payment.

mahmed0715:
Fix #1701 
@jmiridis I have added but I could not test it, can you please test and confirm if we get the invoice_number in back office...

### DIFF
--- a/app/components/Payment/PaymentSelector.native.js
+++ b/app/components/Payment/PaymentSelector.native.js
@@ -152,7 +152,12 @@ class PaymentSelector extends React.Component<{}, { elementFontSize: string }> {
   }
 
   _renderContent(section) {
-    const { currency, context } = this.props;
+    const { currency, context, paymentStatus } = this.props;
+    const donationId = this.props.donationId
+      ? this.props.donationId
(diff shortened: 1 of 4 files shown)


app/components/Payment/PaymentSelector.native.js
```diff
@@ -152,7 +152,12 @@ class PaymentSelector extends React.Component<{}, { elementFontSize: string }> {
   }
 
   _renderContent(section) {
```
norbertschuler: I assume this module is not used in the native apps as we redirect to the browser to do the actual payment (as iOS does not allow this kind of payments within the app). But anyway nice you changed this as well. I just can not run this code.

Issue #1705, opened 2019-12-05
Staging: After initial error automatically submitting the payment for a donation


norbertschuler:
**Describe the bug**
On staging when I donate I get an error first and after I click “try again” the rest of the changes you see in the video happens fully automatic without me clicking anything anymore - even the payment!  

The initial error is caused by a missing donationID either at https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/9e31e98eede839836bb74f2fcb0f546acbcdb46a/app/components/Payment/PaymentSelector.js#L75 or at https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/9e31e98eede839836bb74f2fcb0f546acbcdb46a/app/components/StripePayment/CheckoutForm.js#L105

**Is it on a Production or Development Build?**
staging

**To Reproduce**
1. Go to 'Donate', 
2. Pick our default project 'Campeche Rainforest Restoration'
3. Use the default value of 50 trees

**Expected behavior**
Do the donation process as on production.

**Video**
[autopay.mp4.zip](https://github.com/Plant-for-the-Planet-org/treecounter-app/files/3925835/autopay.mp4.zip)

**Desktop (please complete the following information):**
- OS: latest MacOS
 - Browser latest Chrome
 - Plant-for-the-Planet App Version 1.1.1

mahmed0715:
this is good that it works as the code is made to go the right place if there are some values already set...
Checking the code to fix the actual issue.. @norbertschuler

PR #1714, opened 2019-12-06
Commenting offline payment fixes the issue as we dont use offline any…

mahmed0715:
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1705


norbertschuler:
Using this patch for the web client together with the staging backend works and the bug does not appear any more. When I test it with a native app you get redirected to https://staging.trilliontreecampaign.org/, where the payment is currently still directly done as the fix isn't yet applied. As the API endpoint is also used to deciding which API to use I can not change it to my localhost to test the patch. We should test it again for the native app once we merged this.

PR #1731, opened 2019-12-11
Feature/pledge increment changes

### LINKED ISSUE #1685: PledgeEvents v1.2

harshvitra:
Fixes #1685

Changes in this pull request:
- User can update name, tree count and is Anonymous status
- Lower tree count is not accepted
- New Header added

@sagararyal 

Screenshot_2019-12-11-20-55-27-310_org pftp

### DIFF
--- a/app/components/PledgeEvents/UpdatePledgeEvent.native.js
+++ b/app/components/PledgeEvents/UpdatePledgeEvent.native.js
@@ -103,7 +103,7 @@ class MakePledgeForm extends Component {
         <View>
           <HeaderAnimated
             navigation={this.props.navigation}
-            title={i18n.t('label.increasePledge')}
+            title={i18n.t('label.updatePledgeButton')}
             scrollY={this.state.scrollY}
           />
           <View />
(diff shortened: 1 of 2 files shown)


harshvitra:
@jmiridis 

We are implementing Update pledge page, earlier we had increment pledge where we could only update the tree count. 

**Update pledge page** - here we want the user to be able to change all details except for the email id.

**Increment pledge** api worked in a way where the user had to increase the tree count field, (previous treecount + 1)

**Update pledge page -** here if the user just wants correct a Typo in his name, he will have to add 1 tree to update the Pledge or he wont be able to do it.

**Change needed from your end** - the api to allow users to enter the same tree count along with higher tree count

Issue #1736, opened 2019-12-11
Gift Trees: disabled buttons can still be clicked


norbertschuler:
**Describe the bug**
The gift form now has a disabled buttons (either the small arrow button or the big button with text on it). But a disabled button does only make sense if it is not clickable. Otherwise we do not need a disabled button (UI guidelines!).

**Is it on a Production or Development Build?**
develop

**To Reproduce**
1. Go to 'Gift trees'
2. Click button
3. See error

**Expected behavior**
Button should not be clickable if disabled.

**What actually happens**
Button still can be clicked on.

**Screenshots**
![Screenshot_2019-12-11-22-32-34](https://user-images.githubusercontent.com/1532418/70662894-74805c80-1c67-11ea-89d6-a9769f5facf5.png)

**Smartphone (please complete the following information):**
- Device: Huawei Honor 6
 - OS: Android 6
 - Plant-for-the-Planet App Version 1.1.1

Issue #1735, opened 2019-12-11
Gift Trees: new form does not scroll for keyboard


norbertschuler:
**Describe the bug**
Try to edit a firstname in Gift form. 

**Is it on a Production or Development Build?**
develop

**To Reproduce**
1. Go to 'Gift trees'
2. Click in field for firstname
3. See error

**Expected behavior**
Scroll the layout to always see the field to enter text for.

**What actually happens**
Does not display the input field

**Screenshots**
![Screenshot_2019-12-11-22-31-26](https://user-images.githubusercontent.com/1532418/70662627-ec01bc00-1c66-11ea-951b-709d70767135.png)

**Smartphone (please complete the following information):**
- Device: Huawei Honor 8
 - OS: Android 6
 - Plant-for-the-Planet App Version 1.1.1.1

mahmed0715:
_think we can hide the description text when keyboard is shown. @norbertschuler @sagararyal _

PR #1739, opened 2019-12-12
Disable button click prevent

mahmed0715:
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1736
fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1735

### DIFF
--- a/app/components/GiftTrees/Tabs/FormikFormGift.native.js
+++ b/app/components/GiftTrees/Tabs/FormikFormGift.native.js
@@ -168,7 +168,7 @@ export default class FormikFormGift extends Component {
                         buttonStyles.dualActionButtonTouchable,
                         { top: undefined, bottom: 55, flex: 1, padding: 20 }
                       ]}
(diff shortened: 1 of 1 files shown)


app/components/GiftTrees/Tabs/FormikFormGift.native.js
```diff
@@ -70,9 +70,11 @@ export default class FormikFormGift extends Component {
                   <Text style={styles.add_competition_title}>
                     {i18n.t('label.gift_receipient')}
                   </Text>
-                  <Text style={stylesGift.description}>
-                    {i18n.t('label.gift_trees_description')}
-                  </Text>
+                  {this.state.buttonType === 'next' ? (
```
norbertschuler: Sorry, but hiding the description is not the right solution. Just put this view also into the `KeyboardAwareScrollView` component and it scrolls up with the keyboard as it should do (compare the user gift screen). I will do a patch for this.

PR #1775, opened 2019-12-23
react native map integration

sanjayradadiya:
copy .env.sample and renamed to .env.staging and paste google map API **(googleApiKey)** key there.
Copy and page Google map API key **(googleApiKey)** in local.property inside the Android folder.
 Cope and page google map API **(googleApiKey)** key on the config file.

we need to to run using **`ENVFILE=.env.staging react-native run-android`**  or **`ENVFILE=.env.staging react-native run-ios`**  as google map key is taking from env


norbertschuler:
Does this replace https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1587 and can the older PR get closed?

android/app/build.gradle
```diff
@@ -137,19 +143,20 @@ android {
         applicationId "org.pftp"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 20
-        versionName "1.2.0"
+        versionCode 21
+        versionName "1.2.3"
```
norbertschuler: The version code and name should only be increased if we want to do a new release!

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.5-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.0.1-all.zip
```
norbertschuler: Have you experience with version 6.0.1 of Gradle? Is it recommended to upgrade?

app/actions/apiRouting.js
```diff
@@ -12,7 +12,7 @@ export const getApiRoute = async (routeName, params) => {
   params =
     'api_login_check' === routeName
       ? params
-      : { version: 'v1.1', _locale: locale, ...params };
+      : { version: 'v1.3', _locale: locale, ...params };
```
norbertschuler: Have you talked with @jmiridis about using a newer version of this API call?

app/actions/challengeActions.js
```diff
@@ -18,7 +18,7 @@ export function challenge(challengeDetails) {
     dispatch(setProgressModelState(true));
     return new Promise(function(resolve, reject) {
       let request = postAuthenticatedRequest(route, challengeDetails, {
-        version: 'v1.1'
+        version: 'v1.3'
```
norbertschuler: Same question as above.

app/actions/registerTree.js
```diff
@@ -20,16 +20,21 @@ export function registerTree(
       'plantContribution_post',
       plantContribution,
       {
+        version: 'v1.3',
```
norbertschuler: Same question as above.

app/components/Map/MapboxMap.native.js
```diff
@@ -0,0 +1,870 @@
...
+                    },
+                    poweredContainer: {
+                      display: 'none'
+                    }
+                  }}
+                  debounce={200}
+                  nearbyPlacesAPI="GooglePlacesSearch"
+                  query={{
+                    // available options: https://developers.google.com/places/web-service/autocomplete
+                    key: googleMapApiKey,
```
norbertschuler: The Google Places API is different from Google Maps API, isn’t it. Is the API configured to be used for both?

ios/Podfile
```diff
@@ -74,7 +60,8 @@ target 'TreecounterApp' do
   pod 'Permission-Siri', :path => "#{permissions_path}/Siri.podspec"
   pod 'Permission-SpeechRecognition', :path => "#{permissions_path}/SpeechRecognition.podspec"
   pod 'Permission-StoreKit', :path => "#{permissions_path}/StoreKit.podspec"
-
+pod 'react-native-maps', :path => '../node_modules/react-native-maps'
+pod 'react-native-google-maps', :path => '../node_modules/react-native-maps'
```
norbertschuler: Do these two react native modules not support auto-linking? I‘d like to avoid any manual linking with react-native 0.6x if possible.

app/actions/apiRouting.js
sagararyal: There's no update to challenge so far, so an upgrade is not necessary imo.

app/actions/registerTree.js
sagararyal: Register trees needs newer versions to support geojson.  @jmiridis

app/components/Map/MapboxMap.native.js
sagararyal: @norbertschuler. the API Key is same for both, but I think with Sanjay's setup, the key needs to. be defined both in index.js and .env file.

norbertschuler: @sagararyal .env for iOS, local.properties for Android and config/index.js for JavaScript :-( There should be better solution for the future.

norbertschuler: @sagararyal is the Places API working? In another project I had to setup an billing method for Google first before I could use it even if using some free budget.

sagararyal: @norbertschuler I would prefer to have one place for API Key as well. We can add the API Key in project, if we can secure it or limit it to be run from our domain only.

@sanjayradadiya as we are not using map box, can you rename the file name here?
Please take care of these little cleanups without someone asking.

android/gradle/wrapper/gradle-wrapper.properties
sanjayradadiya: In java v13 gradle 5.5 will not support if you want to use java v13 you need to upgrade gradle to 6.0.1

android/app/build.gradle
norbertschuler: Resolved as @sagararyal reverted this change.

android/gradle/wrapper/gradle-wrapper.properties
norbertschuler: Resolved as I can still also build with Java 8 and Gradle 6.0.1

package.json
```diff
@@ -62,11 +62,13 @@
     "react-native-action-button": "^2.8.5",
     "react-native-check-box": "^2.1.7",
     "react-native-collapsible": "^1.5.1",
+    "react-native-config": "github:luggit/react-native-config#master",
```
norbertschuler: Is `react-native-config` not available as a package?

app/actions/apiRouting.js
norbertschuler: copied to new PR

app/actions/challengeActions.js
norbertschuler: copied to new PR

app/components/Map/MapboxMap.native.js
norbertschuler: If you set your ENVFILE the app/config/index.js can take the value from process.env.googleMapApiKey.

PR #1782, opened 2020-01-01
Hotfix/currency null nan fix

mahmed0715:
@sagararyal currency issue should be fixed now. we can try with this solution,


mahmed0715:
mistake, ignore this.

PR #1785, opened 2020-01-03
Changed Animated header

harshvitra:
Requested by @sagararyal


harshvitra:
The above ESLint error is solved in Welcome Screens PR too

norbertschuler:
I found two issues with this patch:

- [ ] in the new version I can not click the back arrow which worked in the old version
- [ ] in the new version the text flows into the header - the old version what better as not showing that bug:

OLD VERSION:

Simulator Screen Shot - iPhone 11 - 2020-01-10 at 14 39 24

NEW VERSION:

Simulator Screen Shot - iPhone 11 - 2020-01-10 at 14 40 05

PR #1786, opened 2020-01-03
Changed Limit 10

harshvitra:
@sagararyal


norbertschuler:
If this is meant as a real hotfix it should be based onto the master branch (and finished with GitFlow to not forget to merge it into both master and develop branch what GitHub won’t do).

sagararyal:
This doesn't solve anything.

Competition needs to use a pager to load additional competition when then user is scrolling through some competitions.

PR #1793, opened 2020-01-06
Feature/cleaned ui faqs

### LINKED ISSUE #1790: FAQs Cleaning UI
@ankitecd

Change the UI as discussed. 
Also attaching the image here too - 

FAQs

ankitecd:
Fixes #1790

- [x] Implement SafeAreaView

unchecked as the issue with title still exists. – Sagar

----------
@harshvitra Please check


ankitecd:
@harshvitra @sagararyal 

Why do we need Refresh Control on FAQs ? 
How frequently is the data changing ?

sagararyal:
@ankitecd The data might not change often, but since we don't have websocketsc to update the data in user's device when we change data, I think it might make sense to add refresh control. If you think otherwise we can ignore it, and merge this.

norbertschuler:
> Why do we need Refresh Control on FAQs ?

I think that's what we call "Over-Engineering" -> I would not recommend it. There are other places where Refresh Control makes more sense.

harshvitra:
> @ankitecd The data might not change often, but since we don't have websocketsc to update the data in user's device when we change data, I think it might make sense to add refresh control. If you think otherwise we can ignore it, and merge this.

For Refresh control to be executed user will have to scroll down and it is not what user will attempt in the FAQs screen. 

So lets merge this and find another solution for changing data.

harshvitra:
FAQs


@sagararyal removed the purple color and reduced the size of question mark

norbertschuler:
- [ ] Currently this PR is broken and cannot be tested, check the lint errors:
```
/home/circleci/ttc-app/app/components/FAQ/index.native.js
  80:20  error  'textColor' is not defined  no-undef

/home/circleci/ttc-app/app/components/Header/HeaderNew.native.js
  4:10  error  'SafeAreaView' is defined but never used  no-unused-vars
```

norbertschuler:
- [ ] After I defined a `textColor` I could run the code, but the header does not work at all for my iPhone X simulator on iOS13.3 - I can either see the header text nor can I click the back arrow. The SafeAreaView does not work for this.

Simulator Screen Shot - iPhone 11 - 2020-01-10 at 15 09 08

harshvitra:
@sagararyal , this PR was stuck due to header issue

I fixed it in Login and merged it here, please check

PR #1795, opened 2020-01-07
Feature/login change

harshvitra:
Bugs in this push

- [x] Safe Area View?

Screen Shot 2020-01-07 at 3 05 51 PM

- [x] Back Error
Screen Shot 2020-01-07 at 3 06 02 PM

---------
Changes in this Push - 

1. Removed old header
2. Changed Design according Sagar's design
3. Fixed eye icon issue

@sagararyal please check


app/locales/en/loginlabels.json
```diff
@@ -2,6 +2,7 @@
   "reset": "Reset.",
   "dont_have_account": "Don't have an account?",
   "login": "Login",
+  "log-in": "Log In",
```
norbertschuler: Let's use variable name convention for resource keys using only lower letters and underscore ('a-z' and '-') - so please change to `log_in` or just keep the old name as it does not have to be changed if the content changes.

harshvitra: **log_in is already used in redempltionlabels.json -**

"log_in": "Please login or sign up to redeem your code."

sagararyal:
@harshvitra Safe area view issue, still exists. click login from the welcome screen.

app/locales/en/loginlabels.json
norbertschuler: ok, sorry, why don't you just kept the old name ;-)

norbertschuler: ok, sorry, but why do we have to use "Login" and "Log in" within the app. That's not consistent language for the UI, is it?

harshvitra: We are now redesigning UI, so these changes will slowly be consistent everywhere. 

@sagararyal , we have a good point raised by Norbert here, I guess we are using Sign In at some places.

norbertschuler:
> @harshvitra Safe area view issue, still exists. click login from the welcome screen.

Agree: Same wrong header for iPhoneX and iOS13.3 as in https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1793 (cannot see all text and can not click either):

Simulator Screen Shot - iPhone 11 - 2020-01-10 at 15 16 05

harshvitra:
@sagararyal please test now, if this merges it will fix the error in FAQs page too

PR #1799, opened 2020-01-07
Added Flat List in Competitions

harshvitra:
Added Flat List in Competitions to improve loading speed 

@sagararyal please test


sagararyal:
@harshvitra it actually doesn't seem to change the fact that everything loads at once.

Screen Shot 2020-01-07 at 7 08 50 PM

harshvitra:
> @harshvitra it actually doesn't seem to change the fact that everything loads at once.

@sagararyal  I tried its giving unexpected issues, this might need little more work 

My suggestion - We go ahead with Flatlist and Removal of double loader commits for now, and improve app with these small commits

Issue #1826, opened 2020-01-20
Implement Dynamic CDN, Language and Country


sagararyal:
@jmiridis has implemented the following:
https://www.trilliontreecampaign.org/public/v1.0/en/config

# 1
The application will need to retrieve the url for 
cdnMedia
pdfs
and attach it to the media.

In future media wont be accessible through trilliontreecampaign.org domain

# 2
if a user is not logged in, 
set locale = locale
set currency = currency
set country = country

if user signups or is logged in
[this locale, currency, and country, should be overridden by data in user-profile ]

this data can be used to pre fill forms.

mahmed0715:
I will work on this.. @sagararyal do you want to give idea about the location currency selector for mobile app, ?

sagararyal:
@mahmed0715 you could start the PR without changing anything. you'll receive the correct values from API.

We'll show a screen for user to confirm the locale and currency probably after welcome screen. You can already implement part 2 without worrying about the UI.

PR #1828, opened 2020-01-20
Intialization of cdn media urls 

mahmed0715:
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1826
partially working, the urls are not ready yet.


sagararyal:
@mahmed0715 works fine on web.

On Mobile the URLs do not resolve as app does not make api call to /config

Screen Shot 2020-01-20 at 10 14 02 PM

mahmed0715:
@sagararyal The PR is updated with the correct formatted url from api, its good with test backend for now. Jorgo will deploy to prod server.
We can merge I think..

app/selectors/index.js
```diff
@@ -46,6 +47,7 @@ export const pledgesSelector = state => getPledges(state);
 export const postedPledgesSelector = state => getPostedPledges(state);
 export const currenciesSelector = state => getCurrencies(state);
 export const getCurrency = state => getGlobalCurrency(state);
+export const getCdnMediaUrl = state => getCdnMedia(state);
```
norbertschuler: I am new to redux, therefore I would like to ask how is this definition related to the declaration of the function with the same name in `fetchLocation`?

app/components/App/TreeCounter.native.js
```diff
@@ -11,10 +11,12 @@ import { bindActionCreators } from 'redux';
 import { SafeAreaView } from 'react-navigation';
 import { initLocale } from '../../actions/getLocale.native.js';
 import { fetchCurrencies } from '../../actions/currencies';
+import { fetchConfig } from '../../actions/fetchLocation';
 class App extends Component {
   componentDidMount() {
     initLocale();
     this.props.fetchCurrencies();
+    this.props.fetchConfig();
```
norbertschuler: Why do you also load the config within `fetchLocation`? And don’t we need to load the config for the web client?

norbertschuler: I just recognized `fetchLocation` is a function currently only loaded from the web client, so it implicitly load the config. But I would suggest not doing any API calls implicitly within the code of other API calls.

norbertschuler: Changed it to make `fetchConfig` independent from `fetchLocation` for the web client.

app/selectors/index.js
norbertschuler: If I compare this with currencies `getCurrency` is used from this file and there is not local function `getCurrency` in https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/app/actions/currencies.js?

norbertschuler: I think you saved the cdnMedia config with redux, but never read the value again from redux. You just store it as a local variable within `app/actions/fetchLocation.js` and use it from there, right?

mahmed0715: yes. true. i was intended to use redux value but it was not needed for the need, but I think this might be needed next time when we want those values . but kept as is.

sagararyal:
Looks good! test and staging do not work because there are no media in that bucket.

Issue #1848, opened 2020-01-23
TypeError in /donate-trees


Plant-for-the-Planet:
## Error in Plant-for-the-Planet App

**TypeError** in **/donate-trees**
Cannot read property 'getParam' of undefined

[View on Bugsnag](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5e299afab9be07001932c8b3?event_id=5e299afa0055ec77b7b20000&i=gh&m=ci)

## Stacktrace

    http://0.0.0.0:8080/47.js:32 - DonationTreesContainer.componentDidMount

https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/586e19c80843f5f377c5a1f6a08ef981244d1b4e/app/containers/DonateTrees/index.js#L65

[View full stacktrace](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5e299afab9be07001932c8b3?event_id=5e299afa0055ec77b7b20000&i=gh&m=ci)

*Created automatically via Bugsnag*

mahmed0715:
I am on it.

Issue #1846, opened 2020-01-23
TypeError in app/components/PlantProjects/PlantProjectSnippet.native.js:105


Plant-for-the-Planet:
## Error in Plant-for-the-Planet App

**TypeError** in **app/components/PlantProjects/PlantProjectSnippet.native.js:105**
null is not an object (evaluating 'k.taxDeduction')

[View on Bugsnag](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5e2951d508ce58001aeca7ac?event_id=5e2951d50055ec2ffca60000&i=gh&m=ci)

## Stacktrace

    app/components/PlantProjects/PlantProjectSnippet.native.js:105 - value
    app/containers/PublicTreeCounterContainer/index.js:40 - 

[View full stacktrace](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5e2951d508ce58001aeca7ac?event_id=5e2951d50055ec2ffca60000&i=gh&m=ci)

*Created automatically via Bugsnag*

Issue #1843, opened 2020-01-23
TypeError in app/components/DonateTrees/index.native.js:294


Plant-for-the-Planet:
## Error in Plant-for-the-Planet App

**TypeError** in **app/components/DonateTrees/index.native.js:294**
undefined is not an object (evaluating 'c.paymentSetup.countries')

[View on Bugsnag](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5e2918af84cf78001a7ac7a5?event_id=5e2918af0055e740d18c0000&i=gh&m=ci)

## Stacktrace

    app/components/DonateTrees/index.native.js:294 - 

[View full stacktrace](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5e2918af84cf78001a7ac7a5?event_id=5e2918af0055e740d18c0000&i=gh&m=ci)

*Created automatically via Bugsnag*

mahmed0715:
@norbertschuler anyone worked on this already? I had depencency issue thus i was blocked...

norbertschuler:
> @norbertschuler anyone worked on this already? I had depencency issue thus i was blocked...

Nobody as far as I know.

PR #1838, opened 2020-01-21
Develop undefined issue

mahmed0715:
Fixes #1836 ?

@sagararyal pls check

PR #1850, opened 2020-01-23
Fix not existing props values on donate trees

mahmed0715:
@norbertschuler please check Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1848 https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1843 https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1846


norbertschuler:
I can’t say if with this fix all the issues with missing data are fixed as I couldn’t say for https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1838. This is just a patch for this special case which I can test tomorrow, but the software needs a general refactoring and check of all use cases of the donation process in order to be seen as stable again. Currently I regard this part of our software as unstable.

mahmed0715:
just fixing adhoc for now, we need huge refactoring and re engineering the donation process.. @norbertschuler

mahmed0715:
i see the payment failed because of this api https://www.trilliontreecampaign.org/public/v1.3/en/paymentInfo/KW64UGK5XA payment info null

app/components/DonateTrees/index.native.js
```diff
@@ -148,7 +148,9 @@ export default class DonateTrees extends React.PureComponent {
 
   // open your gateway
   openGateWay = async url => {
+    const localUrl = 'http://localhost:8081' + url;
```
sagararyal: @mahmed0715 do we need this on the repo?

mahmed0715: nope, i tried to have the actual url we are redirecting to web, which is not actually debuggable from mobile app.. 
now i removed all those ..
@sagararyal

PR #1854, opened 2020-01-24
reactivated the old loading indicator

### LINKED ISSUE #1837: Progress indicator for long running (API) transaction not working anymore
We seem to have lost our progress dialog (spinner) appearing in the native apps for longer transactions triggered with `dispatch(setProgressModelState(true));` and stopped with `dispatch(setProgressModelState(false));`.

norbertschuler:
Fixes #1837

Should we bring back the old loading indicator for API queries for our native apps?


norbertschuler:
@haideralishah what is your opinion about this? Have you removed the old loading indicator intentionally or accidentally for indicating API calls?

sagararyal:
@norbertschuler I'd recommend to phase out this, and instead use the in tab loader @haideralishah  has implemented.

There are instances where this comes on top of the new loader.

I think we should identify places and Haider can implement it.

PR #1858, opened 2020-01-24
Feature/add environment support

### LINKED ISSUE #1840: Add environment support
Currently we have a file `app/config/index.js` containing the configuration of the app. We lately added a `.env` file containing an API key used in natively Android and iOS and also getting imported into our `app/config/index.js`.

Next steps:
- migrate all the configuration values which vary for different environments into the `.env` file
- add support for using different environment files (e.g. production, development and test) by adding flavors for Android and custom schemes for iOS and change webpack for the web client
- Android flavors and iOS schemes could have different app names, bundle ids and icons to be able to install apps for different environment at the same time on one device
- automatically build different apps for the different environments by CircleCI (best first do #1841)

sanjayradadiya:
Fixes #1840 

Changes in this pull request:
- Added sample .env for develop, staging, production
- Create new config file for react-native index.native.js and For web client old file
- Web client now use index.js config and default .env.develop file

@
To do

- [x] CircleCI support for .env


.gitignore
```diff
@@ -57,5 +57,8 @@ coverage
 #app build
 web/prod
 web/server
-app/config/index.js
-app/config/index.js.env.*
+
+#.env
+.env.production
+.env.develop
+.env.staging
```
norbertschuler: I think this is not necessary as already done with the lines above:
```
# Enviroment
#
.env*
!.env.sample
```

app/config/index.native.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * CREATE A COPY OF THIS FILE AND NAME IT: index.js
+ */
+
+/**
+ * Parameters that are provided by the server to initialize the App.
+ */
+
+import config from 'react-native-config';
```
norbertschuler: If you don't want to use my solution with the `env.native.js` and `env.js` files, please delete them in this patch. Also we can probably can delete `index.js.dist`.

sanjayradadiya: @norbertschuler I am not aware of your solution can you please explain if your solution is better then we will implement it

norbertschuler: @sanjayradadiya my solution isn't better, just different :-) It loads the config values to the process.env array for react-native in order to avoid having to duplicate `index.js` to `index.native.js`:
```javascript
import Config from 'react-native-config';

// copy Config to process.env to use React-Native like React
for (let e in Config) process.env[e] = Config[e];
```

norbertschuler:
- [] `npm run android` does not seem to work anymore as the flavors renamed the Gradle tasks (e.g. `installDevelopDebug` and `installProductionRelease`) and therefore `installDebug` does not exists anymore which the react-native CLI want to invoke by `react-native run-android`.

PR #1870, opened 2020-01-28
Update donate to support saved treecounter in profile

mahmed0715:
@sagararyal for v1.3 we always use saved supportedTreecounter in user profile if its not from support button or link.


mahmed0715:
yes its for both the platforms, web also works good. @sagararyal

PR #1874, opened 2020-01-30
Fixed Missing locale for pledge

harshvitra:
Fixes issue on production - Missing locale for Pledge


norbertschuler:
I already fixed this error in the develop branch with https://github.com/Plant-for-the-Planet-org/treecounter-app/commit/586e19c80843f5f377c5a1f6a08ef981244d1b4e - I think it won't help doing a hotfix in our master branch as this error is only located in the native apps and we would have to do a full release of the iOS and Android app to fix this. So I vote for closing this PR without doing anything and wait for the next regular release of the native apps.

PR #1880, opened 2020-02-03
Reverts menu to v1.1

sagararyal:
The initial idea for the 1.4 update was to remove menu items from the mobile apps; however they are now not dependent on api anymore.

Fixes issue with My Trees not appearing in web.

### DIFF
--- a/app/layouts/sideMenu.js
+++ b/app/layouts/sideMenu.js
@@ -6,7 +6,7 @@ const PublicSideMenuSchema = menuNameParam => {
   return new Observable(observe => {
     getRequest('public_menu_get', {
       category: menuNameParam,
-      version: 'v1.4'
+      version: 'v1.1'
     })
       .then(({ data }) => {
         observe.next(data);
@@ -20,7 +20,7 @@ const AuthenticatedSideMenuSchema = menuNameParam => {
   return new Observable(observe => {
     getAuthenticatedRequest('data_menu_get', {
       category: menuNameParam,
-      version: 'v1.4'
+      version: 'v1.1'
     })
       .then(({ data }) => {
         observe.next(data);


norbertschuler:
@jmiridis Have you prepared anything special in the backend to support a v1.4 version of the menu for the apps? If yes, is it o.k. if we do back to version v1.1 now?

jmiridis:
> 
> @jmiridis Have you prepared anything special in the backend to support a v1.4 version of the menu for the apps? If yes, is it o.k. if we do back to version v1.1 now?

Menu API returns data from DB table, selecting options based on min/max version fields.

image

The version fields assume the last 3 digits to be the minor version, the preceding 2 digits to be the major version, and whatever comes before that to be the version.

Not sure if this answers your question.

norbertschuler:
> > @jmiridis Have you prepared anything special in the backend to support a v1.4 version of the menu for the apps? If yes, is it o.k. if we do back to version v1.1 now?
> 
> Menu API returns data from DB table, selecting options based on min/max version fields.
> 
> The version fields assume the last 3 digits to be the minor version, the preceding 2 digits to be the major version, and whatever comes before that to be the version.
> 
> Not sure if this answers your question.

Thank you. That's good to know as now I can derive the answer myself: 
- For the _web client_: up to version 1.2999 the menu supports the menu item to `/my-trees`, but starting from version 1.3 the menu items is replaces with the route to `/home`
- For the _native apps_: the register trees route is only supported up to version 1.3, not existing afterwards any more

jmiridis:
> 
> 
> > > @jmiridis Have you prepared anything special in the backend to support a v1.4 version of the menu for the apps? If yes, is it o.k. if we do back to version v1.1 now?
> > 
> > 
> > Menu API returns data from DB table, selecting options based on min/max version fields.
> > The version fields assume the last 3 digits to be the minor version, the preceding 2 digits to be the major version, and whatever comes before that to be the version.
> > Not sure if this answers your question.
> 
> Thank you. That's good to know as now I can derive the answer myself:
> 
>     * For the _web client_: up to version 1.2999 the menu supports the menu item to `/my-trees`, but starting from version 1.3 the menu items is replaces with the route to `/home`
> 
>     * For the _native apps_: the register trees route is only supported up to version 1.3, not existing afterwards any more

This means we have to add a new entry (copy) with the route `/home` with different, non-overlapping min/max version

PR #1886, opened 2020-02-03
changed bottom navi to vertically align text and thin font icons

### LINKED ISSUE #1875: Design issues with of new bottom navigation
The new bottom navigation does not seem to align the text at the same base line (Donate and Compete icon). That makes the "Donate" menu item looks like it's hovering (also because it's higher then the rest). Looks like praying to the tree for me ;-)

Simulator Screen Shot - iPhone 11 - 2020-01-30 at 16 12 29

Also the icons now have different thickness of the lines within the icons (Donate and Compete icon) making them looks like they do not really fit together.

norbertschuler:
Fixes #1875

Before:
73660996-91e70b00-4699-11ea-8ebd-6e1956a1372b

After:
73661004-96abbf00-4699-11ea-9af0-0df5058ce383


norbertschuler:
@sagararyal what is your opinion about this patch? I know the icons look smaller, but that's the price to align the text and stroke thickness of the font icons and image icons used in a mix.

PR #1894, opened 2020-02-04
Implemented new tree count selector

mahmed0715:
@sagararyal implemented new ui for treecount selector


norbertschuler:
Is there any issue for this explaining what this patch should do or some design? Then please link in the description.

norbertschuler:
@sagararyal thanks for the context. I remember I did some fixes on this page, but it was difficult and a redesign is a good idea. Maybe that could even help making it more stable.

norbertschuler:
The challenges of the old screen have been:

- try to allow to show very big amounts of money for currencies like "Usbekistan-SUM"
IMG_E7242

- try to allow to enter a big amount of trees to be donated
IMG_E7243

- failed to allow to enter bigger sums a 9999 for a custom sum, which e.g. for "Usbekistan-SUM" did not make very much sense
IMG_E7244

Issue #1919, opened 2020-02-10
Wrong navigation after login / no redirect to expected screen


norbertschuler:
When you use the app without logging in and then opening the "Me" or the "Competition" screen, you see the login screen. After logging in the "Me" screen is displayed, but the bottom navigation has the "World" item selected. (Before the navigation got changed, the "Me" screen was the first and the "World" screen was the last which now has been swapped.)

![Screenshot_2020-02-08-13-16-17](https://user-images.githubusercontent.com/1532418/74085030-4fe10f00-4a75-11ea-90cd-12ab293330ff.png)

One would except the login window is able to redirect the user to the screen he initially wanted to navigate to, but this functionality does not work any more (if it ever has before).

norbertschuler:
The new component [StaticTabbar](https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/app/containers/Menu/StaticTabbar.tsx) does not seem to be able to change it's internal state other then when a user clicks on a tab. We would need the possibility to change the currently selected tab manually through changing the state of this component.

PR #1929, opened 2020-02-10
Redirect to login when user clicks on Competition or Me

harshvitra:
Redirect to login when user clicks on Competition or Me.


norbertschuler:
I made a movie showing the following issues:

- [ ] I did not login directly, but used the app as a guest and navigated to "Competition". First the login form is directly shown. After login you see the new login info page displayed for a second, then the empty login form again for a second and then get redirect to the "Me" screen ("World" tab highlighted).
- [ ] Afterwards I navigated to "Competion" and then to the "Me" screen the layout "jumps" into place.
- [ ] I logged out again and navigated back to "Competition" as a guest where I again get shown the login form. Clicking the back arrow displays the new login info page. I tested the same for the "Me" screen. Does that makes sense?

[login.mp4.zip](https://github.com/Plant-for-the-Planet-org/treecounter-app/files/4184829/login.mp4.zip)

norbertschuler:
- [ ] There are some more places in https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/app/components/Navigators/AppDrawerNavigator.js#L221 where `LoginContainer` is used execpt for "Competition" and "Me". Don't you also want to change them?
- [ ] It would be great to have a functionality to redirect the user to the route where he wanted to navigate to after having logged in as requested in #1919

PR #1937, opened 2020-02-12
Version 1.3.2

sagararyal:
Update to Web, Android & iOS

Release Note:

```
Good day Tree Planters!
We've added support for currency localization, and made few additional improvements to your profile. With you support, over 65 projects plant trees around the world. Where are you planting next?
```


norbertschuler:
Let’s rather use an issue than a PR to plan the next version?

PR #1969, opened 2020-02-17
leaderboard filtring and sorting issue fixed

### LINKED ISSUE #217: Explore - Sorting and filtering by drop down does not work
Explore page - Sorting or filtering by both the drop downs don't work. Nothing gets logged in the Nwtwork tab of the browser.

image

haideralishah:
Fixes #217

Changes in this pull request:
- Filtring is now configured with 1 week, 1 year, and all time.

@sagararyal


app/containers/Leaderboard/index.js
```diff
@@ -205,10 +205,12 @@ class LeaderBoardContainer extends React.Component {
             );
           }
           if (exploreData.timePeriods) {
-            timePeriodsInfo.timePeriods = exploreData.timePeriods;
-            timePeriodsInfo.timePeriodsKeys = Object.keys(
-              timePeriodsInfo.timePeriods
-            );
```
norbertschuler: You do not have to replace this code with `timePeriodsInfo.timePeriodsKeys = ['all', '1w', '1y'];` - the old code retrieving the keys dynamically still works and avoids duplicating the information about the keys in our sources.

app/containers/Leaderboard/index.js
```diff
@@ -205,10 +205,12 @@ class LeaderBoardContainer extends React.Component {
             );
           }
           if (exploreData.timePeriods) {
-            timePeriodsInfo.timePeriods = exploreData.timePeriods;
```
norbertschuler: The fact that the API call https://test.trilliontreecampaign.org/public/v1.1/de/exploreData (without params) returns  `"timePeriods":{"all":"insgesamt","1-year":"letztes Jahr","5-years":"letzte 5 Jahre"}` has to be regarded as a bug in the API and should rather be fixed in the backend as getting patched in our frontend. Now the frontend doesn't use the correct translation from the backend anymore and e.g. forgot to localize these texts into German. 

Can you please discuss with @jmiridis how to solve this best? How have you solved this in the native apps? Have you completely ignored this API call? Maybe @jmiridis can provide a newer version 1.2 of this API call returning the newly supported timePeriods with proper translations?

app/containers/Leaderboard/index.js
haideralishah: `the old code retrieving the keys dynamically still works`
not working properly it always gets the same list because of the wrong parameters
so the correct params should be `['all', '1w', '1y']` these params also use in the mobile app.
kindly suggest me what should I do?

norbertschuler: I meant to write it like this (replacing the first part of the code if you can't make @jmiridis return the right keys from the backend and keeping the second part retrieving the keys as before):
```javascript
timePeriodsInfo.timePeriods = {
  all: 'All Time',
  '1w': 'This Week',
  '1y': 'Last year'
};
timePeriodsInfo.timePeriodsKeys = Object.keys(
  timePeriodsInfo.timePeriods
);
```

norbertschuler: Did it myself.

app/server/formSchemas/explore.js
```diff
@@ -27,7 +27,8 @@ export default {
   },
   timePeriods: {
     all: 'all time',
-    '1-year': 'last year',
+    '1y': 'last year',
+    '1w': 'last week',
     '5-years': 'last 5 years'
   }
```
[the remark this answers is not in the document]
norbertschuler: I reset this as this file is generated by the server and we shouldn't patch anything inside.

app/containers/Leaderboard/index.js
norbertschuler: As @jmiridis probably had no time to adapt the backend API call, I would agree to do it in the client now as it already was done for the native apps. I added the translations (took them from the native app resources already existing).

PR #2006, opened 2020-02-26
Restrict Landscape mode

harshvitra:
Landscape mode is now restricted both on Android and iOS


harshvitra:
> Don't you want to support landscape anymore? I have noticed the new navigation does not work with it. :-)

Not just navigation, other pages as well. 

Once we can work on the designs, we will again enable landscape mode

PR #2014, opened 2020-02-28
separate API url from website

### LINKED ISSUE #1927: Separate URL for API and website in config/environment
In order to be able to host the API at a different location as the website, the application should be changed so the URL for the API calls differ from the URL of the website in the application. You have to figure our where to use the website URL and where to use the API URL in our application. Also please check not getting CORS issues (eventually the backend has to add something to the server configuration).

norbertschuler:
Fixes #1927

I added a new env variable `api_url` for all API calls made (also for the `suggest` API call which I am now using as `suggest.php` due to CORS issues). The other URLs the application is redirecting to are not changed, so the backend where the client is hosted still have to support some URLs and can not just be a stupid web server hosting the client files. E.g. the web server still must be able to serve the URL for doing the donation process in the web browser. Here is a list of places where still `scheme` and `host` are used in the client (most of them in the native app):

Bildschirmfoto 2020-02-28 um 11 18 55

- [x] I also added `api_url` to CircleCI and buddy.works as environment variables.
- [x] Having another URL for the API requires the server hosting the client to correctly set the `access-control-allow-origin` header.

While I still added the same server name in the sample env files, you can change the `api_url` to another server for testing, e.g. the new Heroku backend API url.


norbertschuler:
> but where is `host` being used now?

You can see remaining usages of `host` and `scheme` in my screenshot from my code search above.

@sagararyal if we merge this with the new names of the backends https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/2014/commits/439a7f59e42949f9f0ab05fa21d02ac080527bb9 we should tell everybody these backends are not ready yet (still fixing bugs). And do you also want to add the new production backend in https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/439a7f59e42949f9f0ab05fa21d02ac080527bb9/.env.production.sample or don't we have decided yet about the final name of the production API?

PR #2062, opened 2020-03-23
map project route to donate-tree route in web client

### LINKED ISSUE #2059: Add donate button to single project page on the web
Currently the single project page (after https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/2058 got merged) shown in the web client for the shared project url, e.g. "Plant trees with Plant-for-the-Planet at https://test.trilliontreecampaign.org/project/1", does not display a "**donate**" button. We will have to add that in the future, so the user can donate to a project from the single project page.

Compare the discussion in the original pull request here https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1792#issuecomment-598349859

norbertschuler:
also patches the DonateTreeContainer to support projectSlug as parameter instead of id

Fixes #2059

@sagararyal please check if this works, e.g. with /project/1 or /project/49 - it's some ugly patch I don't like. Hopefully it does not break anything else :-(

- A first alternative solution to this patch would be a rewrite rule in the web server to change `/project/*` to `/donate-trees/*` - but that only works with a webserver, so you cannot test it locally.

- A second alternate solution would be to use a different route for the share link, e.g. /donate-trees/1 which isn't yet supported in the native apps by changing the line https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/394e9d405d0d7d493e625967b586a419a8e67df2/app/components/PlantProjects/PlantProjectFull.native.js#L139  to ```getLocalRoute('app_donateTrees') +```

### DIFF
--- a/app/components/PlantProjects/PlantProjectFull.native.js
+++ b/app/components/PlantProjects/PlantProjectFull.native.js
@@ -136,7 +136,7 @@ class PlantProjectFull extends React.Component {
             context.scheme +
             '://' +
             context.host +
-            getLocalRoute('app_donateTrees') +
+            getLocalRoute('app_selectedProject') +
             '/' +
             this.props.plantProject.id
           }
--- a/app/components/PlantProjects/PlantProjectFull.native.js
+++ b/app/components/PlantProjects/PlantProjectFull.native.js
@@ -136,7 +136,7 @@ class PlantProjectFull extends React.Component {
             context.scheme +
             '://' +
             context.host +
-            getLocalRoute('app_selectedProject') +
+            getLocalRoute('app_donateTrees') +
             '/' +
             this.props.plantProject.id
           }
--- a/app/components/App/TreeCounter.js
+++ b/app/components/App/TreeCounter.js
@@ -17,9 +17,9 @@ import Footer from '../Footer';
 const SelectPlantProjectContainer = lazy(() =>
   import('../../containers/SelectPlantProject')
 );
-const SelectedPlantProjectContainer = lazy(() =>
-  import('./../../containers/SelectedPlantProject')
-);
+//const SelectedPlantProjectContainer = lazy(() =>
+//  import('./../../containers/SelectedPlantProject')
+//);
 const GiftTreesContainer = lazy(() => import('../../containers/GiftTrees'));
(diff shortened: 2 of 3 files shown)


sagararyal:
Empty Page when selecting select different project.

Screen Shot 2020-03-23 at 5 06 35 PM

app/components/PlantProjects/PlantProjectFull.native.js
```diff
@@ -136,7 +136,7 @@ class PlantProjectFull extends React.Component {
             context.scheme +
             '://' +
             context.host +
-            getLocalRoute('app_selectedProject') +
+            getLocalRoute('app_donateTrees') +
```
norbertschuler: The share link link /donate-trees/projectid does not work in the native apps.

norbertschuler:
> Empty Page when selecting select different project.

@sagararyal how did you selected another project? Can you please describe what you did?

app/components/PlantProjects/PlantProjectFull.native.js
norbertschuler: Also this change makes all the other changes of this PR completely unnecessary as well as all the code adding support for the new sharing url /project/projectSlug.

norbertschuler:
> Empty Page when selecting select different project.

I think this bug already existed with the old url e.g. https://www.trilliontreecampaign.org/donate-trees/1

app/components/PlantProjects/PlantProjectFull.native.js
norbertschuler: I reverted this change. @sagararyal if you want to do this, we should dismiss this PR and make your change as a new PR and just a patch directly in the develop branch.

norbertschuler:
After I saw all the issues in #2066 the page https://www.trilliontreecampaign.org/donate-trees/1 still has, I think it also no good idea to use it as a replacement for https://www.trilliontreecampaign.org/project/1, so I would close this PR again.

app/components/PlantProjects/PlantProjectFull.native.js
harshvitra: I had also tried various ways to get this one done

PR #2088, opened 2020-03-31
Update README.md

niteshgrg:
Fixes the missing step in ios readme for setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ To run the app as prod, useful for testing features like (hashed js/css):
   brew install watchman
   npm install -g react-native-cli
   npm install
+  cd ios && pod install
   ```
 
 * Please use node v 12.14.1 LTS; App doesn't build with 13.0 + versions.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,6 +40,7 @@ To run the app as prod, useful for testing features like (hashed js/css):
   brew install watchman
   npm install -g react-native-cli
   npm install
+  cd ios && pod install
```
norbertschuler: Please change to `cd ios && pod install && cd ..`, so the next command can run again on the same directory level as before.

README.md
```diff
@@ -40,6 +40,7 @@ To run the app as prod, useful for testing features like (hashed js/css):
   brew install watchman
   npm install -g react-native-cli
   npm install
```
norbertschuler: Isn't that a general setup necessary for iOS and Android? Who put that just to the iOS setup process?

PR #2164, opened 2020-04-07
Fixed broken multiple trees registration

### LINKED ISSUE #2119: Register (multiple) trees is broken
**Describe the bug**
The form submits, but show’s again without error and no trees are registered. 

production 

**To Reproduce**
1. Go to ‚Register multiple trees'
2. Fill in number of trees, name and choose position 
3. Click ‚register tree'

### LINKED ISSUE #2161: Register - "Add a Tree" Button broken - Easter soon
**After filling out register form, clicking 'Add a Tree' doesn't trigger any events**
This is the case for Single tree and Many Trees.

**This is a bug on the production app. The Web-App doesn't have this issue. iOS I am unsure about.**

**This is important as we expect quite a lot of users to come to the app at Easter (because of Baumgeschenke, spring starting now, holidays, etc.**
Because of the spring part, I just received three different failed registered trees -  all this problem.
And I tried myself ==> same problem.

**To Reproduce**
1. Go to "Me"
2. Click on "Register Tree"
3. Enter information and choose location
4. Press Button: Add a Tree
5. Nothing happens

**Expected behavior**
Either one:
1. A message telling me that my input was wrong
2. It adding the tree(s) to my trees

**Smartphone:**
- Device: Samsung Galaxy S7
- OS: Android 8.0.0
- Samsung Experience version: 9.0
- Plant-for-the-Planet App Version 1.3.2.900027

Thanks for your time!

norbertschuler:
Fixes #2119
Fixed #2161

Changes in this pull request:
- fixed wrong reference to form for registering multiple trees
- changed wording between registering one or multiple trees to submit button
- changed label "TOTAL TREE" for number of trees to "Number of Trees"
- fixed bug in user contribution cards

### DIFF
--- a/app/components/UserContributions/ContributionCard.js
+++ b/app/components/UserContributions/ContributionCard.js
@@ -275,10 +275,9 @@ export default class ContributionCard extends React.Component {
               : null}
             {!isPending ? (
               <TextSpan>
-                {' '}
                 {cardType && cardType.length > 0
-                  ? cardType.charAt(0).toUpperCase() + cardType.slice(1)
-                  : ''}
+                  ? ' ' + cardType.charAt(0).toUpperCase() + cardType.slice(1)
+                  : ' ' }
               </TextSpan>
             ) : null}
             {mayUpdate ? (
@@ -385,10 +384,9 @@ export default class ContributionCard extends React.Component {
             ) : null}
             {!isPending ? (
               <TextSpan>
-                {' '}
                 {cardType && cardType.length > 0
-                  ? cardType.charAt(0).toUpperCase() + cardType.slice(1)
-                  : ''}
+                  ? ' ' + cardType.charAt(0).toUpperCase() + cardType.slice(1)
+                  : ' ' }
               </TextSpan>
             ) : null}
             {mayUpdate ? (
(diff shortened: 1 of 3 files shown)


app/components/UserContributions/ContributionCard.js
```diff
@@ -275,10 +275,9 @@ export default class ContributionCard extends React.Component {
               : null}
             {!isPending ? (
               <TextSpan>
-                {' '}
                 {cardType && cardType.length > 0
-                  ? cardType.charAt(0).toUpperCase() + cardType.slice(1)
-                  : ''}
+                  ? ' ' + cardType.charAt(0).toUpperCase() + cardType.slice(1)
```
sanjayradadiya, line 279: I am not sure which is best let me know 
can we use here template string
so we no need to concate using + sign

```suggestion
                  ? ` ${cardType.charAt(0).toUpperCase()} ${cardType.slice(1)}`
```

norbertschuler: I have no clue what is the better solution, but there shouldn't be spaces between the capital letter the rest of the word or you would see "D onation" and "G ift" in the list :-)

norbertschuler: Maybe the `+` is only doing string concatenation while the other is using a template engine parser? Then I would prefer simple string concatenation, but I am not sure.

PR #2206, opened 2020-04-17
use dedicated api_url for API calls

norbertschuler:
This will activate the complete separation of the backend and frontend by using a different url for the API than the domain the client is hosted. After all backends are deployed this way, we can remove the proxy in the backend routing the different traffic for the client and backend.

Before we merge this, please test the standalone client installation using the production backend at https://standalone.trilliontreecampaign.org/ and also build the native apps with this patch to check if they do not have any problems.


sagararyal:
Android app works fine. (tested the donation flow) redirects to correct pages.
Web seems to work fine as well.

Issue #2240, opened 2020-04-24
Show the project list as is ignoring redux


mahmed0715:
Currently the featured projects list has some issues with redux, we need to avoid using redux to show the list as we get from api.

PR #2241, opened 2020-04-24
Use live data from api call in featured list: web

mahmed0715:
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2240
Please check @sagararyal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,7 @@ coverage
 #app build
 web/prod
 web/server
+
+
+# lock 
+*.lock
--- a/app/components/SelectPlantProject/index.js
+++ b/app/components/SelectPlantProject/index.js
@@ -42,7 +42,7 @@ class SelectPlantProject extends Component {
       expanded: false,
       pageIndex: 0,
       filteredProjects: props.plantProjects,
-      featuredProjects: props.plantProjects,
+      featuredProjects: props.featuredProjects,
       priceSortedProjects: props.plantProjects,
       searchFieldValue: '',
       imageViewMore: false,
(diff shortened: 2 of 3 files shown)


mahmed0715:
Check out this Video of the demo
Check out this video: https://share.vidyard.com/watch/sNYRK66zgYuw4nffKQ2Qxt?

jmiridis:
@sagararyal @norbertschuler could you please approve this PR?

norbertschuler:
Testing with development backend showing "Trees4Trees" first when you want to donate trees with my local testing environment of the PR at http://localhost:8080/donate-trees - I would expect "Yucatán" to be the first project shown from the results of the API call https://app-development.plant-for-the-planet.org/public/v1.1/de/plantProjects/featured - so is that right or wrong now?

Bildschirmfoto 2020-04-29 um 15 34 22

norbertschuler:
Testing with **production** backend showing "Yucatán" first when you want to donate trees with my local testing environment of the PR at http://localhost:8080/donate-trees - but it's the last project returned by the API call Request URL: https://app.plant-for-the-planet.org/public/v1.1/de/plantProjects/featured - I would expect to show the order exactly as returned by the API call and the API call to return "Yucatán" first if we want that to be:

Bildschirmfoto 2020-04-29 um 15 43 42

norbertschuler:
@sagararyal @mahmed0715 Except the fact the last item of the API results got shown first, the other projects are shown in the right order. But I cannot approve this issue if it should show the projects in the order resulting from the API call.

norbertschuler:
I recognized there is a bug in `react-slick` showing the last items first if you use the settings `infinite: true`. It does not happen when you use `infinite: false` :-(

norbertschuler:
I fixed the bug with `react-slick` by rendering the slider elements first before the slider gets rendered like proposed in https://stackoverflow.com/questions/56438595/react-slick-slider-items-are-miss-ordered-first-item-is-rendered-as-last-items.

@sagararyal @jmiridis Now of course the order of the projects on the production backend needs to be changed once we have released the next version of the web client containing this fix (but not before!).

PR #2244, opened 2020-04-25
Feature/GitHub actions

norbertschuler:
This will build a develop version of the native client with GitHub actions resulting into
- an Android APK file
- an iOS IPA file

connected with the development backend


sagararyal:
@norbertschuler we could use curl to upload these apk or ipa to browserstack. 

https://www.browserstack.com/app-live/rest-api

PR #2250, opened 2020-04-28
allow challenge end date to be chosen in the future

### LINKED ISSUE #2245: Future end date for competitions not accepted
The app don't accept an end date in future in creating a challenge.

norbertschuler:
Fixes #2245

- fill the date picker with the former edit value
- changed the setFieldValue property
- do not format the date if not a date object


harshvitra:
I tested the functionality, works as expected. 

just curios, why try-catch for the date function @norbertschuler ?

norbertschuler:
> I tested the functionality, works as expected.
> 
> just curios, why try-catch for the date function @norbertschuler ?

I did that as the formatDateToMySQL function sometimes got called with the already SQL-formatted string instead a date object and otherwise crashed with an exception. Of course it would have been better to find out why it was called this way and eliminating such calls but I couldn’t find out yet, so I just made sure calling the formatting function won’t crash.

PR #2361, opened 2020-05-20
Fixing update to treecounter after redeem

### LINKED ISSUE #2344: Redemption API call response ignored
**Describe the bug**

The code of https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/app/actions/redemptionAction.js completely ignores the response from the API calls and therefore:
- misses to show the proper error or failure messages returned by the API call 
- misses to update the clients state with the resulting `gift` and `treecounter` objects and therefore not adding new user contribution into the client until it got restarted (users say they can not see the trees they just redeemed their vouchers for)

both

**To Reproduce**
1. Go to 'Me' and choose 'Redeem' button
2. Enter a valid code (get examples from backend)
3. Redeem the voucher code (also enter wrong code to see error results)

**Expected behavior**
Check API call response and react on it correctly.

**What actually happens**
Response of API call is ignored.

**Screenshots**
Bildschirmfoto 2020-05-12 um 09 12 45

mahmed0715:
Fixes #2344 
@norbertschuler please check if the update to treecounter is done , I didnt see any issue with error message.
Can you please review.

### DIFF
--- a/app/actions/redemptionAction.js
+++ b/app/actions/redemptionAction.js
@@ -1,5 +1,7 @@
 import { postAuthenticatedRequest } from '../utils/api';
-import { loadUserProfile } from './loadUserProfileAction';
+import { treecounterSchema } from '../schemas/index';
+import { mergeEntities } from '../reducers/entitiesReducer';
+import { normalize } from 'normalizr';
 
 export function validateCodeAction(data, params) {
   return postAuthenticatedRequest('validateCode_post', data, params);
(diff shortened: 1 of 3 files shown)


app/actions/redemptionAction.js
```diff
@@ -1,9 +1,22 @@
...
 
 export function setRedemptionCodeAction(data, params) {
-  return postAuthenticatedRequest('convertCode_post', data, params);
+  return dispatch => {
+    return new Promise(function (resolve, reject) {
+      postAuthenticatedRequest('convertCode_post', data, params)
+        .then((res) => {
+          dispatch(
+            loadUserProfile(null)
+          );
```
norbertschuler, line 15: I rather thought of something like merging the response objects in our existing entities instead of triggering another API call. Have you tested that? Or are these response objects not suitable for merging?

Issue #2257, opened 2020-04-28
The 4th tab of a project on the 'donate trees' page doesn't work


enoralecuyer:
**Describe the bug**
When I click on the 4th tab/circle under the title of a project, it appears blank for 2 seconds, then redirects me to the 3rd tab/circle, which is the donor details.

**Is it on a Production or Development Build?**
Production

**To Reproduce**

1. Go to https://www.trilliontreecampaign.org/donate-trees (without being logged in)
2. Click on the button 'donate trees'
3. Under the 'project' title, click on the 4rd circle.
4. See error with a blank page, before being automatically redirected to 3rd tab/circle.

**Expected behavior**
I expected to see more information about the project, or to not see a 4rd tab/circle.

**What actually happens**
The page appears blank for 2 seconds, then redirect me to the 3rd tab/circle, which is the donor details.

**Screenshots**
![donatetrees](https://user-images.githubusercontent.com/48727972/80532790-82522800-8951-11ea-957e-64bd534e2f46.png)

**Desktop (please complete the following information):**
- OS: Windows 10
 - Browser: Firefox
 - Plant-for-the-Planet App Version: live

**Additional context**
I was not logged in.

sagararyal:
Todo:

- [ ] Disable clicking on the circles, as it seems to be misleading.

Context:
The fourth page in this case gets only populated after a donation is created from the backend. So the user shouldn't be able to reach there without entering the donation details.

mahmed0715:
on it.

PR #2366, opened 2020-05-22
Fix donate slider manual navigation on the dots, removed click handler

mahmed0715:
Fix https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2257
@sagararyal @norbertschuler

### DIFF
--- a/app/components/DonateTrees/index.js
+++ b/app/components/DonateTrees/index.js
@@ -351,7 +351,6 @@ export default class DonateTrees extends Component {
     };
     const settings = {
       dots: true,
-      dotClass: 'disbale-click',
       nextArrow: (
         <NextArrow checkValidation={this.checkValidation} context={this} />
       ),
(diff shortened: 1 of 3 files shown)


app/components/DonateTrees/index.js
```diff
@@ -351,6 +351,7 @@ export default class DonateTrees extends Component {
     };
     const settings = {
       dots: true,
+      dotClass: 'disbale-click',
```
norbertschuler, line 354: Typo? `disable-click`?

norbertschuler: I only found a `dotsClass` in the documentation: https://react-slick.neostack.com/docs/api#dotsClass - so maybe this never really worked and if used correctly could be used to disable the clicking on the dots?

norbertschuler: The documenation says the default value is `slick-dots` which you already use in the CSS, so this line is completely useless and might be deleted.

norbertschuler: Removed that line myself.

PR #2419, opened 2020-06-08
Fix no image issue in leaderboard- country - details

harshvitra:
In Leaderboard - click on a country

Each country has its own leaderboard,
If there are no images there we have to show a Random Image.

### DIFF
--- a/app/components/LeaderboardRefresh/Schools/SchoolsLeaderBoard.js
+++ b/app/components/LeaderboardRefresh/Schools/SchoolsLeaderBoard.js
@@ -21,7 +21,7 @@ import GetRandomImage from '../../../utils/getRandomImage';
 
 const SchoolsLeaderBoard = ({ navigation }) => {
   const [queryresult, setQueryResult] = useState(null);
-  const [period, setPeriod] = useState('1w');
+  const [period, setPeriod] = useState('1y');
   const [orderBy] = useState('planted');
 
   useEffect(() => {
@@ -164,8 +164,8 @@ const CompanyListItem = ({ onPressListItem, item, index }) => {
             }}
           />
         ) : (
-          <GetRandomImage name={item.caption} />
-        )}
+            <GetRandomImage name={item.caption} />
+          )}
       </View>
       <View style={styles.countryBody}>
         <View style={styles.countryNameCont}>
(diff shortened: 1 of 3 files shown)


app/components/LeaderboardRefresh/Schools/SchoolsLeaderBoard.js
```diff
@@ -21,7 +21,7 @@ import GetRandomImage from '../../../utils/getRandomImage';
 
 const SchoolsLeaderBoard = ({ navigation }) => {
   const [queryresult, setQueryResult] = useState(null);
-  const [period, setPeriod] = useState('1w');
+  const [period, setPeriod] = useState('1y');
```
norbertschuler, line 24: Haider has build redundant code here, so please make sure every change you make is done for TPO, Individuals, Company, Schools and maybe Countries (five times!).

sagararyal: @norbertschuler this was changed only for school and TPOs because the data is quite less for 1w.

harshvitra:
> Leader-board shows country flag images for all the countries I could see on the first page across the 3 categories and when you click further on any category, it shows images for some individual accounts and no image for some. When you click "all time" it shows the default image for most accounts. This is for both iOS and Android devices.

Can you share a screenshot, I can see Images.

Chiagozie-Udeh:
> > Leader-board shows country flag images for all the countries I could see on the first page across the 3 categories and when you click further on any category, it shows images for some individual accounts and no image for some. When you click "all time" it shows the default image for most accounts. This is for both iOS and Android devices.
> 
> Can you share a screenshot, I can see Images.


Capture33
Capture333


I hope those help to understand.

norbertschuler:
I can confirm there is no image for Sagar in Leaderboard of Nepal using this patch on develop branch with develop backend and the CDN servers used for develop, but that's because Sagar has no image on the CDN servers used for develop:

Simulator Screen Shot - iPhone 11 - 2020-06-10 at 17 47 34

If I switch to production API, then it seems to work well:

Simulator Screen Shot - iPhone 11 - 2020-06-10 at 17 55 08Simulator Screen Shot - iPhone 11 - 2020-06-10 at 17 55 20

PR #2444, opened 2020-06-10
Feature/heroku web

sarveshwarge:
Fixes #

Changes in this pull request:
- changed app.json for configuring Heroku deployments
-
-

@

### DIFF
--- a/app.json
+++ b/app.json
@@ -23,7 +23,7 @@
     },
     "host": {
       "description": "Website server domain - add app url",
-      "value": ""
+      "required": "false"
     },
     "api_url": {
       "description": "API endpoint",
--- a/app.json
+++ b/app.json
@@ -1,4 +1,73 @@
 {
   "name": "TreecounterApp",
-  "displayName": "TreecounterApp"
+  "displayName": "TreecounterApp",
+  "description": "this repository contains the code of the web clients and the native iOS and Android apps of the Trillion Tree Campaign at https://www.trilliontreecampaign.org/ written with React-Native",
+  "keywords": ["treecounter", "react-native"],
+  "website": "https://www.trilliontreecampaign.org/",
+  "repository": "https://github.com/Plant-for-the-Planet-org/treecounter-app",
+  "logo": "https://www.trilliontreecampaign.org/assets/apple-touch-icon-57x57.png",
+  "success_url": "/",
+  "stack": "heroku-18",
+  "env": {
+    "googleMapApiKey": {
+      "description": "Google Maps Api Key"
+    },
+    "mediaPath": {
+      "description": "Media Path",
+      "value": "/media/cache"
+    },
+    "scheme": {
+      "description": "Website server protocol",
+      "value": "https"
+    },
+    "host": {
+      "description": "Website server domain - add app url",
+      "value": ""
+    },
+    "api_url": {
+      "description": "API endpoint",
+      "value": "https://app-development.plant-for-the-planet.org"
+    },
+    "base": {
+      "description": "API base url. Debug mode off: on: \"/app_dev.php\" (requires login)",
+      "value": "pass"
+    },
+    "debug": {
+      "description": "local console debugging switch",
+      "value": "true"
+    },
+    "currency": {
+      "description": "Currency",
+      "value": "EUR"
+    },
+    "bugsnagApiKey": {
+      "description": "Bug snag API Key"
+    },
+    "androidAppId": {
+      "description": "Android App ID",
+      "value": "org.pftp"
+    },
+    "iosAppId": {
+      "description": "iOS app ID",
+      "value": "1444740626"
+    },
+    "locationApikKey": {
+      "description": "Location API Key"
+    },
+    "env": {
+      "description": "env should be develop, staging, production",
+      "value": "develop"
+    }
+  },
+  "formation": {
+    "web": {
+      "quantity": 1,
+      "size": "free"
+    }
+  },
+  "addons": [
+    {
+      "plan": "papertrail:choklad"
+    }
+  ]
 }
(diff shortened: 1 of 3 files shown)


app.json
```diff
@@ -1,4 +1,77 @@
...
+      "description": "Website server domain - add app url",
+      "required": "false"
+    },
+    "api_url": {
+      "description": "API endpoint",
+      "value": "https://app-development.plant-for-the-planet.org"
+    },
+    "base": {
+      "description": "API base url. Debug mode off: on: \"/app_dev.php\" (requires login)",
+      "required": "false"
```
norbertschuler, line 34: Should be required true and set to empty by default as otherwise the api requests fail.

package.json
```diff
@@ -19,7 +22,8 @@
     "clean": "./clean.sh",
     "ios": "ENVFILE=.env.develop react-native run-ios --scheme 'TreecounterApp(.env.develop)'",
     "android": "ENVFILE=.env.develop react-native run-android --variant 'developDebug'",
-    "metro": "react-native start --reset-cache"
+    "metro": "react-native start --reset-cache",
+    "heroku-postbuild": "echo \"api_url=$api_url\nenv=$env\nscheme=$scheme\nmediaPath=$mediaPath\nhost=$host\nbase=$base\ncurrency=$currency\n\" >> .env.production && webpack --config ./web/config/webpack.prod.config.js --progress --profile --bail"
```
norbertschuler, line 26: Better copy all our specified environment variables into the .env file.

app.json
sarveshwarge: If I set it as "required": "true" then it can't be submitted empty. let's keep it "required": "false" and default value empty. cool?

package.json
sarveshwarge: Done

sarveshwarge: done

package.json
```diff
@@ -19,7 +22,8 @@
     "clean": "./clean.sh",
     "ios": "ENVFILE=.env.develop react-native run-ios --scheme 'TreecounterApp(.env.develop)'",
     "android": "ENVFILE=.env.develop react-native run-android --variant 'developDebug'",
-    "metro": "react-native start --reset-cache"
+    "metro": "react-native start --reset-cache",
+    "heroku-postbuild": "echo \"api_url=$api_url\nenv=$env\nscheme=$scheme\nmediaPath=$mediaPath\nhost=$host\nbase=$base\ncurrency=$currency\ngoogleMapApiKey=$googleMapApiKey\ndebug=$debug\nbugsnagApiKey=$bugsnagApiKey\nandroidAppId=$androidAppId\niosAppId=$iosAppId\nlocationApikKey=$locationApikKey\" >> .env.production && webpack --config ./web/config/webpack.prod.config.js --progress --profile --bail"
```
norbertschuler, line 23: If we want to use source maps for JavaScript debugging, we should use `webpack.server.config.js` instead of `webpack.prod.config.js`.

norbertschuler:
> Also, do we explicitly need to set this? can't it run on the latest version?
> `"engines": { "node": "12.14.1" },`
> 
> Actually I added it while debugging, I think that should be removed.

I am not sure, if using the latest node version 14 instead if 12 is still a problem, but I remember it was a problem last year when you run `npm install` for a new installation (I don't exactly remember which package raised the problem and I still have node 12 installed locally). Maybe you want to test again?

sarveshwarge:
I have node version 14 on my machine and the web version build and worked as expected. 
I am saying that the node engine property was added by me while debugging on Heroku.
you can check this commit - https://github.com/Plant-for-the-Planet-org/treecounter-app/commit/29b15e827b370d4dee9428599069cb5c94b27ef9
so if we know the app works fine even without that. should we remove it?

norbertschuler:
> I have node version 14 on my machine and the web version built and worked as expected.
> I am saying that the node engine property was added by me while debugging on Heroku.
> you can check this commit - [29b15e8](https://github.com/Plant-for-the-Planet-org/treecounter-app/commit/29b15e827b370d4dee9428599069cb5c94b27ef9)
> so if we know the app works fine even without that. should we remove it?

GitHub Action use `v12.18.0` and CircleCI use `v12.14.1` and I personally have `v12.13.1` installed. So I just don't know if it works yet. Perhaps you just test it. In order to be sure, you should have completely delete the `node_modules` folder and any further caches. I just found this warning in one of our scripts: https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/8d5d480da3ec6debc7804e6221a4c9495706e162/clean.sh#L31-L33

package.json
norbertschuler: @sarveshwarge so do we want to have source maps on the review apps or not? This is what source maps look like on https://development.trilliontreecampaign.org/

Bildschirmfoto 2020-06-15 um 13 50 37

sarveshwarge: is it optional? I mean if we use `webpack.server.config.js` and not set the API key will it work? the idea here is we want to define all default env variables in the app.json, and the review app can be deployed automatically without any manual work, like adding additional config vars.

Plant-for-the-Planet: I don't think we need more API keys for using https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/web/config/webpack.server.config.js than for using https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/web/config/webpack.prod.config.js - please compare yourself. It should work.

norbertschuler: I don't think we need more API keys for using https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/web/config/webpack.server.config.js than for using https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/web/config/webpack.prod.config.js - please compare yourself. It should work.

sarveshwarge: Done

PR #2466, opened 2020-06-17
added update screen, wait for loading of config API call

### LINKED ISSUE #2370: Mark outdated app versions and notify the user
We want to mark older app version as outdated. We could use the already existing response from the config API call already existing:

```json
{
  "appVersions": {
    "ios": [

    ],
    "android": [

    ]
  },
  ...
}
```
@jmiridis how is the above information meant to be used inside the apps?

1. We can write some code as following using the above array if it include all valid app versions - but then the backend has to know and return **all** valid versions:
```javascript
if (!appVersions.ios.includes(package.version)) {
   // show the user an information that the app is outdate and a link to the app stores
}
```
2. We can write some code to check if the app version is less then the supported version (so the API call has only to return the lowest supported version and does not need an array):
```javascript
if (package.version < appVersions.ios[0]) {
   // show the user an information that the app is outdate and a link to the app stores
}
```
3. We can send the app version as payload to the config API call to let the backend do the check and afterwards just return the information if the calling app is outdated or not (new boolean flag in the response).

We could use the design from https://github.com/Plant-for-the-Planet-org/treecounter-app/pull/1376 for this.

norbertschuler:
Fixes #2370

Changes in this pull request:
- added update screen from #1376
- wait for loading of config API call in native app (to check outdated app versions)
- increase config API call version to v1.2 (so we can return HTTP status 426 for v1.1 in the future)
- ignore HTTP status 426 API responses

@jmiridis this PR requires backend changes to become active.

### DIFF
--- a/app/components/Menu/index.native.js
+++ b/app/components/Menu/index.native.js
@@ -59,8 +59,7 @@ class Menu extends Component {
     // check for updates
     await this.props.fetchConfig();
     console.log('package version:', version, ' appVersions:', getAppVersions());
-    if (getAppVersions()[Platform.OS] && getAppVersions()[Platform.OS][0] && version < getAppVersions()[Platform.OS][0]) {
-    // if (getAppVersions()[Platform.OS] && getAppVersions()[Platform.OS][0] && version < getAppVersions()[Platform.OS][0]) {
+    if (getAppVersions()[Platform.OS] && version < getAppVersions()[Platform.OS]) {
        // show the user an information that the app is outdate and a link to the app stores
       updateStaticRoute('app_splash_screen', this.props.navigation);
     } else {
--- a/app/components/Menu/index.native.js
+++ b/app/components/Menu/index.native.js
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
 import { View, ScrollView, SafeAreaView, Text, Linking, Platform } from 'react-native';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
 import { debug } from '../../debug';
 import styles from '../../styles/menu.native';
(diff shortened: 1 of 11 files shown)


app/components/Menu/index.native.js
```diff
@@ -55,15 +55,25 @@ export default class Menu extends Component {
...
-      // }
-      updateRoute('welcome_screen', this.props.navigation, 0);
+
+    // check for updates
+    await this.props.fetchConfig();
+    console.log('package version:', version, ' appVersions:', getAppVersions());
+    if (getAppVersions()[Platform.OS] && version < getAppVersions()[Platform.OS]) {
+       // show the user an information that the app is outdate and a link to the app stores
+      updateStaticRoute('app_splash_screen', this.props.navigation);
+    } else {
```
harshvitra, line 65: this.props.fetchConfig().then(
 if else 
);

harshvitra: The fetchConfig function is promise check below diagram - 

promises

app/styles/splashScreen.native.js
```diff
@@ -0,0 +1,53 @@
...
+export default EStyleSheet.create({
+  updateMainView: {
+    flex: 1,
+    backgroundColor: '#fff',
+    justifyContent: 'center',
+    alignItems: 'center'
+  },
+  updateTitle: {
+    fontSize: 27,
+    fontWeight: '800',
```
harshvitra, line 12: Let me change the fonts to OpenSans.

app/components/Menu/index.native.js
norbertschuler: Thank you, that's definitely a better idea as using the `await` statement here.

norbertschuler:
@harshvitra I did you change request. Can you please review again.

Issue #2418, opened 2020-06-08
TypeError in /donate-trees - Cannot read property 'plantProjectPager' of undefined


Plant-for-the-Planet:
## Error in Plant-for-the-Planet App

**TypeError** in **/donate-trees**
Cannot read property 'plantProjectPager' of undefined

[View on Bugsnag](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5ede505e838a3a0018a6f595?event_id=5ede505e005b6d3b32540000&i=gh&m=ci)

## Stacktrace

    app/actions/loadTposAction.js:33 - 

[View full stacktrace](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5ede505e838a3a0018a6f595?event_id=5ede505e005b6d3b32540000&i=gh&m=ci)

*Created automatically via Bugsnag*

norbertschuler:
<img width="626" alt="Bildschirmfoto 2020-06-10 um 13 54 06" src="https://user-images.githubusercontent.com/1532418/84236464-df7cf400-ab21-11ea-9b5c-8ac8a6d1296f.png">

norbertschuler:
Hopefully fixes with next release.

Issue #2329, opened 2020-05-09
Error in /donate-trees/1 - Unexpected input given to normalize. Expected type to be "object", found "string".


Plant-for-the-Planet:
## Error in Plant-for-the-Planet App

**Error** in **/donate-trees/1**
Unexpected input given to normalize. Expected type to be "object", found "string".

[View on Bugsnag](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5eb65370cb1cb90017fd493a?event_id=5eb65370005b3790a32e0000&i=gh&m=ci)

## Stacktrace

    app/actions/loadTposAction.js:62 - 

[View full stacktrace](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5eb65370cb1cb90017fd493a?event_id=5eb65370005b3790a32e0000&i=gh&m=ci)

*Created automatically via Bugsnag*

norbertschuler:
Problem got raised at May 17th:

<img width="1260" alt="Bildschirmfoto 2020-05-22 um 09 34 25" src="https://user-images.githubusercontent.com/1532418/82625492-76cfe500-9c0f-11ea-98c3-aba8562d4542.png">

norbertschuler:
I wonder if the `plantProjectSchema` and all the other schemas defined in https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/develop/app/schemas/index.js still match the data model of the backend or are already outdated:

https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/069898284b7eb4d77c0582080df3af6f89e59d0c/app/schemas/index.js#L53-L57

norbertschuler:
Hopefully fixes with next release.

norbertschuler:
Hopefully fixes with next release.

PR #2483, opened 2020-06-23
check if the API response contains valid data object

### LINKED ISSUE #2480: Refactor check of valid responses from API requests
A lot of errors are throws by not checking for correct result of the API requests, e.g.:

- https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2456
- https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2418 P1
- https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2355
- https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2349
- https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2329 P1
- https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2204
- https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2202

Therefore we should add more checks to the API request code before we use the resulting data in any further modules where they result into all kind of errors.

norbertschuler:
Fixes #2480

Changes in this pull request:
- check if the API response contains a valid data object and throwing an error otherwise

As I generally added this check I will have to make sure:
- [x] Is every API call really returning a data object?
- [x] What happens if the API call throws an error? Is the error always handled correctly if no response data object exists?

### DIFF
--- a/app/actions/apiRouting.js
+++ b/app/actions/apiRouting.js
@@ -1,6 +1,6 @@
 const routes = require('../server/routes/fos_js_routes.json');
 import Routing from './router.min.js';
-import { debug } from '../debug';
+//import { debug } from '../debug';
 import { context } from '../config';
 import { getCdnMediaUrl, getWebMapIds } from './fetchConfig';
 import { getLocaleAsync } from './getLocale';
@@ -9,7 +9,6 @@ Routing.setRoutingData(routes);
 export const getApiRoute = async (routeName, params) => {
   const { api_url, base: baseUrl } = context;
   let locale = await getLocaleAsync();
-  debug('-------------------- Api calling with ', locale);
   const serverName = `${api_url}`;
   params =
     'api_login_check' === routeName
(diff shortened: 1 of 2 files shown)


norbertschuler:
@mahmed0715 do you think this could be helpful for #2418 and #2329?

jmiridis:
There is at least 1 endpoint that does not return an object, just a 200 with an empty body. The deletion of an account.

```
@Rest\Delete("/api/{version}/{_locale}/profileDelete/{userProfile}", name="profile_delete")
```

norbertschuler:
> There is at least 1 endpoint that does not return an object, just a 200 with an empty body. The deletion of an account. 
> ```
> @Rest\Delete("/api/{version}/{_locale}/profileDelete/{userProfile}", name="profile_delete")
> ```
@jmiridis Thank you for this information. This API call is returning a `204 DELETE` status, so I have just added an exception for this status not checking for a valid data in the response. Are there any other API calls not returning a data object. How can I best check this?

PR #2609, opened 2020-08-13
changed logos on launch screen for iOS

norbertschuler:
Fixes broken screen on old iPhone5 or iPhone SE (1st Generation):

File

Changes in this pull request:
- change the order of the logos next to each other instead of on top of each other

@sagararyal please have a look - I would also do it for the Treemapper app, but I cannot apply it on Android that easily.


norbertschuler:
Result on this for me on various iOS simulators:

- iPhone SE 
iPhone SE

- iPhone SE (2nd Gen)
iPhone SE(2nd Generation)

- iPhone X 
iPhone X

- iPhone 11 
iPhone 11

Issue #2475, opened 2020-06-22
Change Tcomb to Formik forms


norbertschuler:
Finish changing all forms from using Tcomb to Formik.

PR #2625, opened 2020-08-17
remove ref from for get the data in TargetContainer

### LINKED ISSUE #2216: undefined is not an object (evaluating 'u.refs.targetContainer.refs')

**TypeError** in **MainActivity**
undefined is not an object (evaluating 'u.refs.targetContainer.refs')

[View on Bugsnag](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5e9daf73e77d2200199794fa?event_id=5e9daf73005a22aec2620000&i=gh&m=ci)

## Stacktrace

    app/containers/TargetContainer/index.js:23 - onSubmitTarget
    app/components/Common/Button/PrimaryButton.native.js:15 - onPress

[View full stacktrace](https://app.bugsnag.com/plant-for-the-planet/plant-for-the-planet-app/errors/5e9daf73e77d2200199794fa?event_id=5e9daf73005a22aec2620000&i=gh&m=ci)

sanjayradadiya:
Fixes #2216

Changes in this pull request:
- undefined is not an object (evaluating 'u.refs.targetContainer.refs')

### DIFF
--- a/app/components/Target/index.js
+++ b/app/components/Target/index.js
@@ -41,6 +41,7 @@ export default class Target extends Component {
       label: countTarget !== 0 ? i18n.t('label.update') : i18n.t('label.save'),
       initialValues: { countTarget, targetYear, targetComment }
     };
+    this.formRef=React.createRef();
   }
   shouldComponentUpdate(nextProps) {
     return JSON.stringify(nextProps) !== JSON.stringify(this.props);
@@ -52,12 +53,12 @@ export default class Target extends Component {
         <TextHeading>{i18n.t('label.set_target')}</TextHeading>
         <CardLayout>
           <TCombForm
-            ref="setTargetForm"
+            ref={this.formRef}
             type={targetFormSchema}
             options={allSchemaOptions}
             value={this.state.initialValues}
           />
-          <PrimaryButton onClick={this.props.onSubmitTarget}>
+          <PrimaryButton onClick={()=>this.props.onSubmitTarget(this.formRef.current)}>
             {this.state.label}
           </PrimaryButton>
         </CardLayout>
--- a/app/components/Target/index.native.js
+++ b/app/components/Target/index.native.js
@@ -20,6 +20,7 @@ export default class Target extends Component {
     this.state = {
       label: countTarget !== 0 ? i18n.t('label.update') : i18n.t('label.save')
     };
+    this.formRef=React.createRef();
   }
 
   shouldComponentUpdate(nextProps) {
@@ -37,14 +38,14 @@ export default class Target extends Component {
             style={{ flex: 1, marginTop: Platform.OS === 'ios' ? 140 : 100 }}
           >
             <Form
-              ref={'setTargetForm'}
+              ref={this.formRef}
               type={targetFormSchema}
               options={this.props.schemaOptions}
               value={this.props.treecounter}
             />
           </CardLayout>
           <PrimaryButton
-            onClick={this.props.onSubmitTarget}
+            onClick={()=>this.props.onSubmitTarget(this.formRef.current)}
             buttonStyle={{ marginTop: 10 }}
           >
             {this.state.label}
--- a/app/containers/TargetContainer/index.js
+++ b/app/containers/TargetContainer/index.js
@@ -17,14 +17,8 @@ class TargetContainer extends React.Component {
     };
     this.onSubmitTarget = this.onSubmitTarget.bind(this);
   }
-  onSubmitTarget = () => {
-    /* debug(
-      '\x1b[45m targetCOntainer',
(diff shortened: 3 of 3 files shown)


app/components/Target/index.js
```diff
@@ -41,6 +41,7 @@ export default class Target extends Component {
       label: countTarget !== 0 ? i18n.t('label.update') : i18n.t('label.save'),
       initialValues: { countTarget, targetYear, targetComment }
     };
+    this.formRef=React.createRef();
```
norbertschuler, line 44: As I am looking for a universal solution we can use throughout our whole codebase I would like to ask what is the difference of this solution vs. https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/4ab1b4d3c5e6cda169b10dd886530bab97d8a836/app/components/Authentication/SignUp/SignUp.js#L145-L162 - can you maybe try to develop one template I can then use within the whole codebase to use refs correctly?

app/components/Target/index.native.js
```diff
@@ -20,6 +20,7 @@ export default class Target extends Component {
     this.state = {
       label: countTarget !== 0 ? i18n.t('label.update') : i18n.t('label.save')
     };
+    this.formRef=React.createRef();
```
norbertschuler, line 23: The same question for the native apps implementation as for the web client. Can you give me an example implementation I can use everywhere for the native apps? What is the difference of this patch against https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/4ab1b4d3c5e6cda169b10dd886530bab97d8a836/app/components/Authentication/SignUp/SignUp.native.js#L44-L50 plus https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/4ab1b4d3c5e6cda169b10dd886530bab97d8a836/app/components/Authentication/SignUp/SignUp.native.js#L84-L91?

norbertschuler (commented):
Can you give a tip how one can reproduce the bug within the local development environment (web client or native apps)?

app/components/Target/index.native.js
```diff
@@ -37,14 +38,14 @@ export default class Target extends Component {
...
           >
             <Form
-              ref={'setTargetForm'}
+              ref={this.formRef}
               type={targetFormSchema}
               options={this.props.schemaOptions}
               value={this.props.treecounter}
             />
           </CardLayout>
           <PrimaryButton
```
norbertschuler, line 47: Btw. this is one example of a form where the ActivityIndicator within the PrimaryButton is not yet integrated (compare https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/1837), so the user does not know if the API requests for updating his target is still working or the app is crashed ;-)

norbertschuler (approved):
Technically it works for me.

app/components/Target/index.js
divsbhalala: we are using TCombForm therefor it will only allow to access the method using ref. if we need to optimise this then we need to change TCombForm to Formik form that we are using in tree registration

app/components/Target/index.native.js
divsbhalala: in container file we are using ref to access the tcombform data, I removed that container ref and pass the tcombform ref to the callback function.

divsbhalala:
> Can you give a tip how one can reproduce the bug within the local development environment (web client or native apps)?

I am also not able to reproduce but based on issue description I make some changes

app/components/Target/index.native.js
sanjayradadiya: we are using TCombForm therefor it will only allow to access the method using ref. if we need to optimise this then we need to change TCombForm to Formik form that we are using in tree registration

app/components/Target/index.js
sanjayradadiya: in container file we are using ref to access the tcombform ref.
I just remove that ref and added tcombform ref in callBack function

sanjayradadiya:
> Can you give a tip how one can reproduce the bug within the local development environment (web client or native apps)?

Actually I am also not able to reproduce this, but issue description I made some changes.

app/components/Target/index.native.js
norbertschuler: The refactoring work to change all remaining old Tcomb form to Formik is also still open in https://github.com/Plant-for-the-Planet-org/treecounter-app/issues/2475 :-)

app/components/Target/index.js
norbertschuler: Ok, then I assume this patch should be the template for doing it everywhere else now.

norbertschuler:
@sanjayradadiya if you mark this draft as ready for review I could approve it.

norbertschuler:
won't be done any more as project will be deprecated

PR #2953, opened 2021-02-22
added refresh token for auth0

ankitecd:
Changes in this pull request:
- Added refresh token and validation of token using same

@norbertschuler please check


norbertschuler (approved):
Works well for me. I set the value of `auth0_token_expires` to zero and the app successfully retrieved a new refresh token for the user.

PR #3077, opened 2021-08-02
disabled Pledge events in native app home screen

### LINKED ISSUE #1471: Add deep link for pledge events
**Describe the feature you'd like to see included in Plant-for-the-Planet App.**
We want to be able to open the native apps (iOS and Android) with an URL as https://staging.trilliontreecampaign.org/pledge/esri-user-conference - therefore we need to add a deep link /pledge/* to the app where * is the slug of the pledge event.

**What are the use cases for this feature?**
Directly open the apps on the phone if a user click on an URL as seen above.

**Examples of this feature or functionality.**
https://staging.trilliontreecampaign.org/pledge/esri-user-conference

**Additional information**
I already starting some work on this issue in the branch https://github.com/Plant-for-the-Planet-org/treecounter-app/commits/feature/deep_link_pledge, but I had to stop as I found out the app cannot just be opened and directly navigated to a pledge event as the component then would fail as it is missing the project and the pledge event data which normally is loaded before. So the code opening the deep-link would need to preload all missing data before it correctly works.

norbertschuler:
Fix #1471 (as not needed any more)

Changes in this pull request:
- disabled Pledge events in native app home screen

### DIFF
--- a/app/components/TreecounterGraphics/Trillion.native.js
+++ b/app/components/TreecounterGraphics/Trillion.native.js
@@ -228,7 +228,7 @@ class Trillion extends PureComponent {
                 {/* Trillion Tree Events Title Ended */}
 
                 {/* Featured events horizontal ScrollView */}
-                {this.props.pledgeEvents &&
+                {/* {this.props.pledgeEvents &&
                   this.props.pledgeEvents.pledgeEvents ? (
                     <View>
                       <ScrollView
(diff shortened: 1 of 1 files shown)


norbertschuler:
@sagararyal I have no more test data for this. Did you have an account with pledge event on staging, so you can check if they disappeared now?

PR #3091, opened 2021-08-12
let the auth token expire 5 minutes earlier to force an early renewal

### LINKED ISSUE #3090: Check Auth0 token expiry time before starting web view with payment
**Describe the feature you'd like to see included in Plant-for-the-Planet App.**

Check if the Auth0 token for the user is valid for more the 5 minutes before starting an authorized web session to make a donation with donate.plant-for-the-planet.org. Either warn the user of renew the token if possible!

norbertschuler:
Fixes #3090

Changes in this pull request:
- let the auth token expire 5 minutes earlier to force an early renewal

### DIFF
--- a/app/utils/user.js
+++ b/app/utils/user.js
@@ -71,7 +71,8 @@ const refreshTokenIfExpired = async () => {
  */
 const tokenIsExpired = async (isAuth0 = false) => {
   let tokenExpiry = await getTokenExpires(isAuth0);
-  let expired = getCurrentUnixTimestamp() > tokenExpiry;
+  // let the token expire 5 minutes before it actually expires
+  let expired = (getCurrentUnixTimestamp() + 300) > tokenExpiry;
   return expired;
 };
 


norbertschuler:
@tejassonar maybe you can look at the code and tell me if the logic it correct. It's a little bit difficult to test this, so if we can just approve the logic is correct, we do not have to test this in real life. (Normally you would write some unit tests for the small helper functions in this file.)

tejassonar:
> @tejassonar maybe you can look at the code and tell me if the logic is correct. It's a little bit difficult to test this, so if we can just approve the logic is correct, we do not have to test this in real life. (Normally you would write some unit tests for the small helper functions in this file.)

Okay. I'll check. Also, I am working on recycling tokens in the planet pay. For that, I have changed the expiration time to 300 seconds. So maybe we can test this as well. So, according to this logic, the app will get a new access token every time they go on planet pay.

norbertschuler:
> For that, I have changed the expiration time to 300 seconds. So maybe we can test this as well. So, according to this logic, the app will get a new access token every time they go on planet pay.

With this configuration the app will try to renew the token for every single API call (as the check for an expired token is done every time the token is requested) :-)

norbertschuler:
The token got renewed on every API call successfully. So I think this proves that it works. I even measure the difference with `console.log("token", tokenExpiry - (getCurrentUnixTimestamp() + 300), tokenExpiry - (getCurrentUnixTimestamp()));` as following:

Bildschirmfoto 2021-08-13 um 16 42 31

